PR #3, opened 2015-11-16
Rewrite :tada:

solnic:
### Description

This is a complete rewrite of the library using a different approach which includes:

* Simple and extendible predicate system
* `Rule` interface on top of predicates
* Rule composition API
* A simple DSL that is only a front-end for rule composition

### TODO

- [x] add predicate system
- [x] add rule system on top of predicates
- [x] add a simple DSL for composing rules
- [x] introduce error objects that can carry data needed for user-facing error messages
- [x] add support for multiple rules with disjunction ie `value.empty? | value.email?`
- [x] add support for multiple rules applied to a value separately
- [x] add support for rules that are applied to collections ie `key(:tasks).each { |value| value.string? }`
- [x] update README


lib/dry/validation/error.rb
```diff
@@ -0,0 +1,43 @@
...
+        include Enumerable
+
+        attr_reader :errors
+
+        def initialize
+          @errors = []
+        end
+
+        def each(&block)
+          errors.each(&block)
```
AMHOL: Does this need `return to_enum unless block_given?` ?

solnic: would not hurt :)

PR #7, opened 2015-11-20
Add Dry::Validation::PredicateSet and BuiltIn

AMHOL:
Relies on https://github.com/dryrb/dry-container/pull/7 but could be done without if necessary

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,6 @@ source 'https://rubygems.org'
 
 gemspec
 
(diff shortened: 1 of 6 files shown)


solnic:
Sweet. Can we just merge in the singleton feature and push a release of dry-container?

solnic:
I updated it to work with the freshly released dry-container which includes `merge` interface and switched back to non-singleton registration.

PR #23, opened 2015-12-02
Refactor schema/validation result objects

solnic:
This PR breaks API but it makes it cleaner:

* `Schema` always uses injected error compiler which can do whatever
* `Schema#messages` **is gone**
* To get compiled messages you can do `Schema#call(input).messages`
* `Schema#call` returns a schema result object which has access to:
  * all validation results (can return successes and failures separately too)
  * an error set
  * compiled messages using the error compiler provided by the schema which uses the error set to produce different representation of raw errors (and it's lazy)
  * potentially coerced params
* `Validation::Result` is now an object gathering all rule results (instances of `Rule::Result`), it provides lower level APIs for working with success and failure results
* `Error::Set` is no longer mutable (yay!) and we simply pass it into `Schema::Result`

This is the first step towards having more error compilers, like I18n, raw hash etc.

### DIFF
--- a/lib/dry/validation/error.rb
+++ b/lib/dry/validation/error.rb
@@ -6,8 +6,8 @@ module Dry
 
         attr_reader :errors
 
-        def initialize
-          @errors = []
+        def initialize(errors)
+          @errors = errors
         end
 
         def each(&block)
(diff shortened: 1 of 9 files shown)


AMHOL:
:+1: LGTM :smile:

PR #24, opened 2015-12-02
I18n and Messages API refactor

solnic:


solnic:
Look up paths will be as follows:

* `namespace.errors.attributes.rule_name.predicate_name.predicate_arg_type`
* `namespace.errors.attributes.rule_name.predicate_name`
* `namespace.errors.predicate_name.predicate_arg_type`
* `namespace.errors.predicate_name`
* `errors.attributes.rule_name.predicate_name.predicate_arg_type`
* `errors.attributes.rule_name.predicate_name`
* `errors.predicate_name.predicate_arg_type`
* `errors.predicate_name`

Does this make sense? Notice that we don't need `base` concept, as we always rely on a specific rule name (which often matches name of the attribute). This makes me think we could drop `attributes` from those paths. Hmmm

/cc @AMHOL @kwando @Bartuz

AMHOL:
@solnic any chance you could update to indicate which are template vars?

Also, do we need `{attribute_name}` in there? So we can do custom messages per attribute like:

```ruby
user:
  attributes:
    address:
      postcode:
        inclusion?: We currently only accept registrations from the United Kingdom.
```

Couldn't think of a decent example :laughing:

solnic:
@AMHOL `{attribute_name}` is not needed because typically `{rule_name}` corresponds to it. ie when you write `key(:postcode)` you actually create a rule called `:postcode` with the predicate `:key?` which argument is set to `:postcode`

AMHOL:
Nice :+1: :smiley:

solnic:
I consider that done. Lemme know if you see something weird. Also, if somebody could tell me how to dynamically merge built-in error messages into I18m locales that would be awesome.

solnic:
:tada: merged

Issue #5, opened 2015-11-19
Add support for reading attribute values


solnic:
Add support for `Rule::Attr` that will be created from `attr(:name)` DSL, the same as `Rule::Key` but uses attr reader rather than [] to access the value.

SunnyMagadan:
Hi, @solnic and @AMHOL 
I would like to contribute into your project. I like the idea and see it like a clean and promising replacement for activemodel validations. I've already started playing around with this feature and made some progress there. Please let me know if you are interested in any help here.

AMHOL:
@SunnyMagadan had a look at your branch and it all looks good to me, if you submit a pull request @solnic can take a look/merge it. Always good to see people taking an interest/helping out, thanks. :smile:

SunnyMagadan:
:bow: Great! Keep doing.

SunnyMagadan:
@solnic I have one question regarding this feature. How do you plan to deal with `optional`? Right now it uses `Rule::Key` with `Implication` under the hood https://github.com/dryrb/dry-validation/blob/master/lib/dry/validation/schema/definition.rb#L10. Also `InputTypeCompiler` uses `:key` for visit `and` `or` `implication` https://github.com/dryrb/dry-validation/blob/master/lib/dry/validation/input_type_compiler.rb#L57. We need to distinguish optional key and optional attr somehow.

solnic:
@SunnyMagadan there is no such thing as an optional attribute, an optional key means a hash may, or may not include a given key. When we talk about validating a model, it means it does have attribute readers for specified attributes. In this case, saying `attr(:foo)` means a model must implement `foo` reader. You can just skip this for attributes, since an attr reader will return `nil` even when the ivar was never set, so one could easily write `attr(:foo) { |foo| foo.none? | foo.int? }` etc.

solnic:
BTW feel free to open a PR early, even when it's not finished. We can collaborate on changsets easily etc :)

SunnyMagadan:
True. It doesn't make sense to have optional attributes. Thank you. I'll open WIP PR soon. :)
Also, I've found that this feature requires changes in `dry-data`, because current version of `InputTypeCompiler` based on it and uses schema conversion into `hash` type. I see that dry-data supports struct, but it requires to define a separate class for this. This is not very efficient for cases when we need to define schema on the fly.

solnic:
[...]

PR #29, opened 2015-12-08
Add support for validation of attributes values

SunnyMagadan:
* Added attr to schema DSL

https://github.com/dryrb/dry-validation/issues/5

### DIFF
--- a/lib/dry/validation/hint_compiler.rb
+++ b/lib/dry/validation/hint_compiler.rb
@@ -61,6 +61,11 @@ module Dry
         name
       end
 
+      def visit_attr(node)
+        name, _ = node
+        name
+      end
+
       def method_missing(name, *args)
         nil
       end
--- a/config/errors.yml
+++ b/config/errors.yml
@@ -36,6 +36,8 @@ en:
 
     key?: "%{name} is missing"
 
+    attr?: "%{name} is missing"
+
     lt?: "%{name} must be less than %{num} (%{value} was given)"
 
     lteq?: "%{name} must be less than or equal to %{num}"
--- a/lib/dry/validation/error_compiler.rb
+++ b/lib/dry/validation/error_compiler.rb
@@ -41,6 +41,11 @@ module Dry
         visit(predicate, value, name)
       end
 
+      def visit_attr(rule, name, value)
+        _, predicate = rule
+        visit(predicate, value, name)
+      end
+
       def visit_val(rule, name, value)
         name, predicate = rule
         visit(predicate, value, name)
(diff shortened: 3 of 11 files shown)


SunnyMagadan:
WIP

solnic:
Awesome :)

lib/dry/validation/rule/attr.rb
```diff
@@ -0,0 +1,21 @@
...
+      def type
+        :attr
+      end
+
+      def evaluate_input(input)
+        input.public_send(name)
+      end
+
+      def call(input)
+        Validation::Rule::Result::EvaluatedValue.new(input, predicate.(input), self)
```
SunnyMagadan: @solnic What do you think about this idea of delayed input evaluation?

solnic: How would that work? I remember we discussed this but now I can't find it for some reason :(

SunnyMagadan: We discussed it here: https://github.com/dryrb/dry-validation/pull/32#discussion_r48425294
It will work in a way that it does `input.public_send(name)` call only when `attr?` predicate returns success, so we will prevent `NoMethodError` to be triggered when `input` doesn't respond to specific attribute expected in schema.
Here is the place where I do this check and call attribute:  https://github.com/dryrb/dry-validation/pull/29/files#diff-f8751eafe5dec4fd9321589ed3e2b792R68

lib/dry/validation/hint_compiler.rb
```diff
@@ -61,6 +61,11 @@ def visit_key(node)
         name
       end
 
+      def visit_attr(node)
+        name, _ = node
```
AMHOL: Think it would make more sense to just return `node.first` here and in `visit_key`

SunnyMagadan: I would leave it as is just for sake of readability. It's not obvious what is hidden under the `node.first` at the first look.

AMHOL: Fair point :+1:

solnic: That was my intention. Although I wasn't super consistent with that tbh.
Gotta change it in some places

SunnyMagadan:
@solnic, @AMHOL it looks like I've added everything that was needed for this feature.
I have only one point that I wasn't sure about. https://github.com/dryrb/dry-validation/pull/29#discussion-diff-49164712R17
I will rebase this PR against fresh master, but I am not going to add anything new here. So, WIP can be removed.
Please check this PR again and let me know if you have any objections, suggestions, questions and so on.
P.S. It might be helpful to have corresponding labels like wip, ready-for-review instead of editing title for PR.

solnic:
We gotta move new rules to [dry-logic](https://github.com/dryrb/dry-logic) and here we only need the DSL part. That's a copy/paste + rename, I hope you won't hate me for this extraction ;) I had to do it because the core system is something that is shared between dry-data and dry-validation now :)

SunnyMagadan:
No problem. I can help you with movement of this PR into dry-logic.

solnic:
Fantastic, thank you!

SunnyMagadan:
Here is corresponding PR for dry-logic https://github.com/dryrb/dry-logic/pull/1

SunnyMagadan:
@solnic I've rebased this PR according to recent changes in dry-logic.

solnic:
I rebased your branch and merged it into master. It's gonna be included in 0.6.0. Thank you <3

SunnyMagadan:
Awesome. Thank you @solnic :cake:

PR #84, opened 2016-03-27
Remove threadsafe, add concurrent

joevandyk:


joevandyk:
I think this is the last usage of ThreadSafe in the dry-rb projects

lib/dry/validation/hint_compiler.rb
```diff
@@ -24,7 +24,7 @@ class HintCompiler < ErrorCompiler::Input
       EXCLUDED = [:none?, :filled?, :key?].freeze
 
       def self.cache
-        @cache ||= ThreadSafe::Cache.new
+        @cache ||= Concurrent::Map.new
```
joevandyk: should `require 'concurrent'` be in this file?

lib/dry/validation/messages/abstract.rb
```diff
@@ -1,5 +1,5 @@
 require 'pathname'
-require 'thread_safe/cache'
+require 'concurrent'
```
solnic: would it be possible to only require `concurrent/map`?

joevandyk: done

solnic:
Thank you :smile:

PR #86, opened 2016-03-29
Add support for enum dry types

coop:
Implements https://github.com/dry-rb/dry-validation/issues/83

This is pretty rough and introduces some duplication but I wanted to put it out there to get some feedback. Happy to make whatever changes are required.

### DIFF
--- a/lib/dry/validation/schema/rule.rb
+++ b/lib/dry/validation/schema/rule.rb
@@ -111,7 +111,7 @@ module Dry
             if predicate.is_a?(::Symbol)
               [target.type, [name, [:predicate, [predicate, args]]]]
             elsif predicate.respond_to?(:rule)
-              [target.type, [name, predicate.rule.to_ast]]
+              [target.type, [name, [:type, predicate]]]
             else
               [target.type, [name, predicate.to_ast]]
             end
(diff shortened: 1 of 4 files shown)


solnic:
Thanks for looking into implementing it. We should actually introduce an interface for dumping entire type into an ast format that can be translated by dry-validation into its rules. That would be the final solution. A quick-win would be to just do `if predicate.respond_to?(:rule); predicate.rule.to_ast; end` but that would be a nasty smell that I'd prefer to get rid of quickly.

I've got a hunch that it would be simple to do the right thing™ already though :)

We could use a trivial ast format for types that looks like this:

``` ruby
[:type, type_object]
```
and extend `SchemaCompiler` in dry-v to be able to visit it:

``` ruby
def visit_type(type)
  # this could return an empty array by default, and only constrained types would dump their rules
  type.rule_ast
end
```

lib/dry/validation/schema/rule.rb
```diff
@@ -110,8 +110,8 @@ def key(predicate, args = [])
           node =
             if predicate.is_a?(::Symbol)
               [target.type, [name, [:predicate, [predicate, args]]]]
-            elsif predicate.is_a?(Types::Constrained)
-              [target.type, [name, predicate.rule.to_ast]]
+            elsif predicate.respond_to?(:rule)
```
coop: This obviously isn't the best solution but @solnic will take a look a reimplementing when support for custom coercions is added.

coop:
I believe I need to add `visit_type` to the hint class. I will add a failing spec.

solnic:
oh, you're right, otherwise we won't be inferring hints from type-based rules. it would have to dump rule to ast and visit it.

PR #87, opened 2016-03-29
Support inferring Dry::Type::Struct validations

coop:
Implements https://github.com/dry-rb/dry-validation/issues/85

Builds on https://github.com/dry-rb/dry-validation/pull/86 - I will rebase when merged.

### DIFF
--- a/lib/dry/validation/schema.rb
+++ b/lib/dry/validation/schema.rb
@@ -48,7 +48,13 @@ module Dry
       def self.create_class(target, other = nil, &block)
         klass =
           if other
-            Class.new(other.class)
+            if other < Types::Struct
+              Validation.Schema(parent: target, build: false) do
+                other.schema.each { |attr, type| key(attr).required(type) }
+              end
+            else
+              Class.new(other.class)
+            end
           else
             Validation.Schema(target.schema_class, parent: target, build: false, &block)
           end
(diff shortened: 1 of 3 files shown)


spec/integration/schema/using_types_spec.rb
```diff
@@ -3,27 +3,56 @@
...
+    end
+
+    it 'handles nested structs' do
+      expect(nested_schema.(person: {name: {given_name: 'Tim', family_name: 'Cooper'}})).to be_success
+    end
+
+    it 'fails when input is not valid' do
+      expect(schema.(person: nil)).not_to be_success
+    end
+  end
```
solnic: you could expand the spec to make sure it works with "nested" structs too

lib/dry/validation/schema.rb
```diff
@@ -48,7 +48,13 @@ def self.new(rules = config.rules, **options)
       def self.create_class(target, other = nil, &block)
         klass =
           if other
-            Class.new(other.class)
+            if other < Types::Struct
+              Validation.Schema(build: false) do
+                other.schema.each { |attr, type| key(attr).required(type) }
```
solnic: looks like we need to call `create_class` here when type is another struct :)

lib/dry/validation/schema.rb
```diff
@@ -48,7 +48,13 @@ def self.new(rules = config.rules, **options)
       def self.create_class(target, other = nil, &block)
         klass =
           if other
-            Class.new(other.class)
+            if other < Types::Struct
+              Validation.Schema(build: false) do
```
solnic: this needs `target: target` otherwise schema's "path" won't be set, which we use to determine where nested structure is located in the input.

PR #88, opened 2016-03-30
Fix empty predicate error message

jodosha:
```ruby
require 'bundler/setup'
require 'dry-validation'

UserSchema = Dry::Validation.Schema do
  key(:sample) { empty? }
end

result = UserSchema.({
  sample: [23]
})

unless result.success?
  puts result.messages
end

# :sample=>["cannot be empty"]}
```

That reads bad: "Sample cannot be empty", but it should be "Sample must be empty".

### DIFF
--- a/config/errors.yml
+++ b/config/errors.yml
@@ -2,7 +2,7 @@ en:
   errors:
     array?: "must be an array"
 
-    empty?: "cannot be empty"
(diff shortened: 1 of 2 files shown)


solnic:
thanks Luca, you made it just in time for 0.7.3 :laughing:

PR #89, opened 2016-03-30
Add JSON input processor

coop:
Working on adding support for https://github.com/dry-rb/dry-types/issues/59.


coop:
This is waiting on `JSON::*` for dry-types. I have a branch that works, just need to tidy up and push. Should be able to do that this weekend.

PR #90, opened 2016-04-01
Confusing/ambiguous aspects of DSL syntax

endash:
As an admitted heretic who initially spurned dry-validation—but has since taken a second look at it after my own effort began to look like an attempt at re-implementing dry-v—I come bearing a n00b's perspective on the syntax, and a change suggestion.

The current syntax—while clearly documented—is moderately confusing, and induces a rather muddled mental model that can persist even after scrutinizing the docs and examples. This can be put down to a number of things:

1.  The left = key, right = value divide is not readily apparent, and in fact can be the opposite of obvious, e.g. `key(:foo).required` *does* require the key, but as a consequence of using `key` and not `optional`.
2.  The required-ness of `key` is implicit, so it doesn't show up as a contrasting element in relation to `optional` (or counter-balancing `.required`) in examples, to help construct the mental model. The only thing that contrasts is `required`, which relates to values.
3.  It goes without saying, but `key`/`optional` are not symmetrical terms or semantically opposed to one another so it's not as obvious that they're *meant* to have opposite/related effect as it could be
4.  On the value side, `maybe` and `required` aren't symmetrical either (granted, a truly symmetrical relationship would be "can't be nil/must be nil", but that'd be a bit pedantic). Their relation to their effects are meaningful, but w/r/t to each other they differ semantically—you wouldn't for instance ever say "This thing is required, but this thing is maybe."
5.  The sides cross-pollinate. `required` and `optional` are a natural pair, but they're used to mean two different things, weakening simultaneously their semantic power AND the right/left distinction. To a lesser extent, "maybe" can also be ambiguous and could easily be interpreted as meaning there maybe a key (with a value, nil or not), rather than the value may be nil. 

My proposal is simply to make things symmetrical and explicit, so that the distinctions being made are clearer even to a naive observer. It also removes most of the ambiguity in the terms used. Misunderstandings are still possible, but most would be resolved with just the left/right rule being pointed out.

```ruby
schema = Dry::Validation.Schema do
  required(:email).not_nil(format?: EMAIL_REGEX)
  optional(:age).maybe(:int?, gt?: 18)
end
```

1.  `key` is now `required`. `required/optional` are opposites, and it surfaces the previously implicit requirement.
2.  `required` is now `not_nil`. It pays to be explicit, here... not only is there `key(:foo).required` confusion to avoid, but the entire question of whether nil is a value or not (and thus satisfying a requirement to have a value) is intractable and a distraction. Rather than make assumptions, we instead make our intent unambiguous and explicit.
3.  `maybe` remains a truncated form of `maybe_nil`, for brevity. The semantics are still slightly off (`mustn't_be_nil` seemed superfluous), but "may be nil" and "not nil" are close enough and relate well to each other regardless.

*Update*: It's now clear to me that `required` does more than ensure the value isn't nil—further semantics I wasn't aware of. How to address that is covered in the discussion.

### DIFF
--- a/examples/basic.rb
+++ b/examples/basic.rb
@@ -1,9 +1,9 @@
 require 'dry-validation'
 
 schema = Dry::Validation.Schema do
-  key(:email).required
+  required(:email).not_nil
 
-  key(:age).required(:int?, gt?: 18)
+  required(:age).not_nil(:int?, gt?: 18)
 end
 
 errors = schema.call(email: 'jane@doe.org', age: 19).messages
--- a/examples/each.rb
+++ b/examples/each.rb
@@ -2,7 +2,7 @@ require 'byebug'
 require 'dry-validation'
 
 schema = Dry::Validation.Schema do
-  key(:phone_numbers).each(:str?)
+  required(:phone_numbers).each(:str?)
 end
 
 errors = schema.call(phone_numbers: '').messages
--- a/examples/form.rb
+++ b/examples/form.rb
@@ -1,9 +1,9 @@
 require 'dry-validation'
 
 schema = Dry::Validation.Form do
-  key(:email).required
+  required(:email).not_nil
 
-  key(:age).required(:int?, gt?: 18)
+  required(:age).not_nil(:int?, gt?: 18)
 end
 
 errors = schema.call('email' => '', 'age' => '18').messages
--- a/examples/nested.rb
+++ b/examples/nested.rb
@@ -1,14 +1,14 @@
 require 'dry-validation'
 
 schema = Dry::Validation.Schema do
-  key(:address).schema do
-    key(:city).required(min_size?: 3)
+  required(:address).schema do
+    required(:city).not_nil(min_size?: 3)
 
-    key(:street).required
+    required(:street).not_nil
 
-    key(:country).schema do
-      key(:name).required
-      key(:code).required
+    required(:country).schema do
+      required(:name).not_nil
+      required(:code).not_nil
     end
   end
 end
--- a/lib/dry/validation/schema/dsl.rb
+++ b/lib/dry/validation/schema/dsl.rb
@@ -23,9 +23,10 @@ module Dry
         end
         alias_method :to_s, :inspect
 
-        def key(name, &block)
+        def required(name, &block)
           define(name, Key, &block)
         end
+        alias_method :key, :required
 
         def optional(name, &block)
           define(name, Key, :then, &block)
--- a/lib/dry/validation/schema/rule.rb
+++ b/lib/dry/validation/schema/rule.rb
@@ -24,11 +24,12 @@ module Dry
           add_rule(rule)
         end
 
-        def required(*predicates)
+        def not_nil(*predicates)
           rule = ([key(:filled?)] + infer_predicates(predicates)).reduce(:and)
 
           add_rule(__send__(type, rule))
         end
+        alias_method :required, :not_nil
 
         def maybe(*predicates)
           rule =
--- a/spec/integration/attr_spec.rb
+++ b/spec/integration/attr_spec.rb
@@ -2,7 +2,7 @@ RSpec.describe Dry::Validation::Schema, 'defining attr-based schema' do
   describe 'with a flat structure' do
     subject(:schema) do
       Dry::Validation.Schema do
-        attr(:email).required
+        attr(:email).not_nil
         attr(:age) { none? | (int? & gt?(18)) }
       end
     end
@@ -22,18 +22,18 @@ RSpec.describe Dry::Validation::Schema, 'defining attr-based schema' do
   describe 'with nested structures' do
     subject(:schema) do
       Dry::Validation.Schema do
-        attr(:email).required
+        attr(:email).not_nil
 
         attr(:age) { none? | (int? & gt?(18)) }
 
         attr(:address) do
           attr(:city) { min_size?(3) }
 
-          attr(:street).required
+          attr(:street).not_nil
 
           attr(:country) do
-            attr(:name).required
-            attr(:code).required
+            attr(:name).not_nil
+            attr(:code).not_nil
           end
         end
 
--- a/spec/integration/custom_error_messages_spec.rb
+++ b/spec/integration/custom_error_messages_spec.rb
@@ -18,7 +18,7 @@ RSpec.describe Dry::Validation do
           config.messages_file = SPEC_ROOT.join('fixtures/locales/en.yml')
         end
 
-        key(:email, &:filled?)
+        required(:email, &:filled?)
       end
     end
 
@@ -38,7 +38,7 @@ RSpec.describe Dry::Validation do
             config.messages = :i18n
           end
 
-          key(:email, &:filled?)
+          required(:email, &:filled?)
         end
       end
 
--- a/spec/integration/custom_predicates_spec.rb
+++ b/spec/integration/custom_predicates_spec.rb
@@ -30,7 +30,7 @@ RSpec.describe Dry::Validation do
           config.predicates = Test::Predicates
         end
 
-        key(:email) { filled? & email? }
+        required(:email) { filled? & email? }
       end
     end
 
@@ -58,7 +58,7 @@ RSpec.describe Dry::Validation do
           end
         end
 
-        key(:email) { filled? & email? }
+        required(:email) { filled? & email? }
       end
     end
 
@@ -68,7 +68,7 @@ RSpec.describe Dry::Validation do
   describe 'custom predicate which requires an arbitrary dependency' do
     subject(:schema) do
       Dry::Validation.Schema(base_class) do
-        key(:email).required(:email?)
+        required(:email).not_nil(:email?)
 
         configure do
           option :email_check
@@ -97,7 +97,7 @@ RSpec.describe Dry::Validation do
         end
       end
 
-      key(:email).required(:email?)
+      required(:email).not_nil(:email?)
     end
 
     expect { schema.(email: 'foo').messages }.to raise_error(
--- a/spec/integration/hints_spec.rb
+++ b/spec/integration/hints_spec.rb
@@ -18,7 +18,7 @@ RSpec.describe 'Validation hints' do
   context 'with yaml messages' do
     subject(:schema) do
       Dry::Validation.Schema do
-        key(:age) do |age|
+        required(:age) do |age|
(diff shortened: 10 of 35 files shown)


fran-worley:
May I offer an alternate opinion?

I personally like: `key(:attr).required` as it is clear. I agree that the use of `maybe` as the opposite of `required` doesn't quite work. From memory, `maybe` was used because `optional` was already being used to describe a key being optional and maybe was the next best thing.

I'm nervous of using `not_nil` instead of required as this doesn't accurately describe what is happening. `required` simply calls the `filled?` predicate which is actually testing that a value isn't empty rather than isn't nil. An important distinction to make particularly in forms where values are not always coerced to nil.

**An alternate suggestion...**

What you could do instead is to loose the `key` vs `optional` declaration altogether and in its place just use key. Any options passed to they key method are applied to the key, any methods called on key are applied to the value. This frees you up to rename maybe to optional, solving the confusion...

 e.g.

```ruby
schema = Dry::Validation.Schema do
  #REQUIRED KEYS

  #both the key and the value are required
  key(:email).required(format?: EMAIL_REGEX)

  #the key is required but no the value.
  key(:age).optional(:int?, gt?: 18) 

  #-------------------------------------------------------------
  #OPTIONAL KEYS

  # the key is optional, but if present value is required
  key(:email, optional: true).required(format?: EMAIL_REGEX) 

  #the key and the value are optional
  key(:age, optional: true).optional(:int?, gt?: 18)
end
```

endash:
> not_nil instead of required as this doesn't accurately describe what is happening.

It describes it better than `required`, since `required` builds in the assumption that `nil` does not satisfy a requirement for a value. For a project supposedly about being explicit and precise, that's a problem, which `maybe` just compounds by bringing in FP-style implications. The fact is, there's a lot of wibbly-wobbly around `nil` and the best way to handle that is to avoid vague terms that may be interpreted in multiple ways. 

> I personally like: key(:attr).required as it is clear. 

That is objectively not the case. It is not clear: it can easily be read as if the key is being made required, when it is the value being made required. You're totally free to prefer the old syntax, of course, but just because you are personally familiar with it doesn't mean it is actually clear, or that it helps present a clear mental model of what is going on.

That's common when discussing API—folks who are familiar with the syntax and all the implicit assumptions and thinking behind it have trouble seeing the ambiguity and unnecessary confusion. It's perfectly understandable, and everyone is free to prefer whatever syntax they want. Clarity is not on the side of the current syntax, though, one way or another.

Now, "not_nil" presents some 20% use case issues—but in the majority of cases it's what is meant, intended, and being done. It may be that other use cases need to be split off in order to improve the accuracy. Splitting it into `not_nil` and `not_empty` (as a DSL, much preferred to `filled?` which implies a value judgment, filled is not the opposite of empty) is a small price to pay. Simply leaving it as one big pile of ambiguity, though, would be ill-advised. You can align the API with what users are actually doing and what actually needs to be communicated, or you can have a confusing, mixed bag syntax that requires going over the documentation several times to internalize.

fran-worley:
I take your point about the lack of objectivity in my comment. 

Would something like this be more explicit?

```ruby
schema = Dry::Validation.Schema do
  #REQUIRED KEYS

  #both the key and the value are required
  key(:email).value(filled?: true, format?: EMAIL_REGEX)

  #the key is required but not the value. Additional validations are only called when the key is filled.
  key(:age).when(:filled?).value(int?: true, gt?: 18) #when could take either predicate names as symbols
  key(:age).when{ true }.value(int?: true, gt?:18) #or a block

  key(:age).when.not(:none?).value(int?: true, gt?: 18) #in active record rails 4 style, you could also support `when.not`
  #-------------------------------------------------------------
  #OPTIONAL KEYS

  # the key is optional, but if present value is required
  key(:email, optional: true).value(filled?: true, format?: EMAIL_REGEX) 

  #the key and the value are only validated when filled
  key(:age, optional: true).when(:filled?).value(int?: true, gt?:18)
end
```

fran-worley:
> Splitting it into not_nil and not_empty (as a DSL, much preferred to filled? which implies a value judgement, filled is not the opposite of empty) is a small price to pay. Simply leaving it as one big pile of ambiguity, though, would be ill-advised.

I guess this isn't dissimilar to the discussions held over the rails method: `present?`. 

With my suggestion above, you could avoid the ambiguity if you wished by doing:

```ruby
  key(:age).value(none?: false, int?: true, gt?: 18)
  key(:email).value(empty?: false)
```
Leaving the `filled?` predicate for those who wish to use it.

endash:
I don't see how that's an improvement over my suggestion, other than it maintains a "combined" `#requirement.` How is `value(none?: false)` An improvement over `not_nil` or `not_empty`?

IMO the fact that `nil` and an empty array are treated as if they were equivalent is a further problem, but I'd have no problems leaving the existing predicates for those who prefer them. I simply think they're idiosyncratic and muddle the syntax for no reason.

fran-worley:
> I don't see how that's an improvement over my suggestion, other than it maintains a "combined" #requirement. How is value(none?: false) An improvement over not_nil or not_empty?

In may ways it isn't, but I had assumed that the biggest issue was with the word `maybe` which has always struck me as a very ambiguous. My suggestion would eliminate it.

I now that I've read your message again I can see that there probably is a margin to add `not_nil?` and `not_empty?` predicates for those who prefer the additional clarity than is offered by `filled?`. 

> IMO the fact that nil and an empty array are treated as if they were equivalent is a further problem, but I'd have no problems leaving the existing predicates for those who prefer them. I simply think they're idiosyncratic and muddle the syntax for no reason.

If you were to remove the `nil` from https://github.com/dry-rb/dry-logic/blob/master/lib/dry/logic/predicates.rb#L33 then those of us who choose to assume that `nil` and `""` are empty would be forced to call 2 methods rather than one. 

One way to address this would be to define the predicates explicitly, and then provide a macro for those of us who want to treat `"", [], nil` as empty and the opposite for filled.

endash:
>  for those who prefer the additional clarity than is offered by filled?

While I definitely personally prefer the added clarity, I do want to take pains to separate the issue from personal preference—my aim is to make the API clearer for my own use, yes, but also to resolve real sources of initial confusion. Simply adding aliases or new predicates doesn't do much for me, if the same broken mental model stays in place for people who are just glancing at the project or are brand new to the project. The project is well documented—reading through it a couple times and it's easy to get a handle on everything. I'm just trying to eliminate some own-goals, if possible, that make that path bumpier than it needs to be.

> One way to address this would be to define the predicates explicitly, and then provide a macro for those of us who want to treat "", [], nil as empty and the opposite for filled.

This explains why @solnic seemed to be dancing around the nil-naming thing when I talked with him this morning... this isn't a trivial concern and I'm not 100% sure of the best way to handle it. I will say first that simply having unambiguous predicates is a start. I also agree that the current `required` behaviour shouldn't require the use of two predicates. At the same time, there's the whole mental model thing, and I can't think of a good name for a combined predicate ATM, either.

Given that the nil situation could use the unambiguous macros/predicates either way, it stands to reason that early examples and intro docs could use the plainly clear `not_nil`, to help ensure the proper mental model, and simply introduce the combined logic along with the other macros.  It's usually best to start with the general case and introduce helpers around more specific cases so this wouldn't be unusual or problematic IMO.

endash:
As a point of clarification... my reading of the source is that `maybe` becomes the `none?` predicate... which is just a test for nil. That introduces some more semantic problems, since `required` takes string/array emptiness into account but `maybe` is all about nil. It's another small thing that makes sense at a local level, but not when viewed as a whole. Having had this discussion I was still under the impression that `maybe` mirrored `required` (I say still, because I was under that impression before, however inaccurate it was). That's not to say that simply flipping it so they *are* perfectly matched makes more sense, however much it cleans up the mental model. What those edge cases are and how to resolve them are very much open questions.

solnic:
I was just thinking about this a little and it seems like `value` for values would be the best option:

``` ruby
required(:age).value(:int?, gt?: 18)

required(:email).value(:filled?)`

required(:address).value(:hash?, :filled?)

optional(:delivery_address).value(:hash?)
```

This reads really well, I think. Oh and `maybe` would stay like it is now.

solnic:
I should add that `value(:filled?)` should be covered by a new macro. One option would be this: `filled(min_size?: 10)`.

endash:
@solnic To confirm that I'm reading this correctly, `value` has the `not_nil` semantics, correct? I think that'll make it a lot more cohesive, particularly as it would mirror `maybe`. I still don't have an alternative for current `required/filled` but I vow to find something to at least consider.

One slight alternative to at least ponder for a moment is `has_value`:

```ruby
required(:first_name).has_value(:string?)
```

The benefit there is it prevents the instinct to associate the `required/optional` with the `value` and is explicit about the fact that it is insisting on a value.

AMHOL:
Do we need to consider `required_key` vs `required_attr` etc?

endash:
I'd hate to add `_key` just to distinguish it from `attr` if key is the 80% use case. I haven't seen `attr` used in any examples or introductory materials and a cursory look at the docs doesn't surface it very readily. How are optional attrs currently handled, since `optional` is an optional key?

At worst, I'd say keep `required` and `optional` for key, and maybe add `required_attr`, `optional_attr` to parallel.

solnic:
No such thing as an optional attr. For model schemas we can create a
special schema type and configure it to use attr rules for required method.
I would also like to move model schemas to a new lib. I am not interested
in maintaining this feature tbf.

daniels:
If you're looking for symmetrical terms wouldn't `is`/`maybe` read better than `has_value`/`maybe`? (Still reads more like "not nil" than "filled" though. But just add a bang and you're set. :wink:)

endash:
@daniels it needs to read without any additional predicates. E.g., right now you can do `key(:name).required` so that'd be `required(:key).is` which doesn't communicate any semantics at all. The semantics need to be somewhat symmetrical, not perfect. "has value" and "maybe [has value]" reflect each other just fine, IMO.

daniels:
@endash Ah, good call there. What about `required(:name).really`. (It reads even better with a bang ...) :stuck_out_tongue_winking_eye:

solnic:
How about `key(:name).orly?` ;)

solnic:
Re `value` vs `has_value` I really prefer the former. In my head, it is immediately obvious that since you specify value with predicates then you really expect a value to be there. I should probably mention that `value` w/o args should raise argument error.

endash:
I'm closing this  because we've just devolved into a what-words-people-like discussion, which IMO doesn't have much point.

PR #91, opened 2016-04-04
Validator explodes when not given a hash when expecting one.

kwando:
A failing test which should probably work.

I've not added any expectations.

### DIFF
--- a/spec/integration/schema/array_schema_spec.rb
+++ b/spec/integration/schema/array_schema_spec.rb
@@ -20,4 +20,8 @@ RSpec.describe Dry::Validation::Schema, 'for an array' do
       1 => { prefix: ["must be filled"] }
     )
   end
+
+  it 'does not explode if array does not contain hashes' do
+    schema.(['hello', 'world'])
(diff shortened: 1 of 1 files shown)


solnic:
If you doesn't trust input type just wrap you rules in a `hash?` block

kwando:
hmm, okey.

solnic:
It's an equivalent of a situation where your code calls `user.valid?` where user is nil or something. It's your responsibility to not let that happen. So ie if you use dry-v against params from a form submission, then you can assume it will be a hash, but if you want to be paranoid you can just do `hash? do .. end` inside your schema, this would protect you from unexpected requests; however, I would suggest to split validation into 2 schemas, one that checks input structure, and return 'invalid request' if it's structure is malformed, and another one to actually apply your business rules. This is a very flexible system, so use it in a way that makes sense for your use cases. I'm especially interested in providing good support for 2-step validation like I just described, it's still being discussed (and I'm not the only one who thinks it's a good idea).

PR #97, opened 2016-04-06
Add support for defining custom input processors

coop:


solnic:
Did you need it just for a custom type to be used? Would a custom type being passed to `key` be enough?

coop:
I don't actually have this problem (which might be enough of a reason not
to merge it) but while building JSON schema I thought it would be nice for
people to be able to override particular types based their requirements.

solnic:
OK so let's hold off with this but keep it open. We should add features when somebody actually wants to use them and we can agree that it's a good idea.

sled:
I really like this idea, there are coercions which are dependent on external parameters (e.g date formats can depend on locale) or come in mangled form like rails' date_(1i), date_(2i) date select helper.

Often you want to write a general schema with basic types such as ints, bools, dates without any input format specific coercions.

An example might be a schema for a person:

```
PersonSchema = Dry::Validation.Schema do
  key(:name).required(:string?)
  key(:date_of_birth).required(:date?)
end
```

This is a very generic schema. In my case I have different input sources to fill this schema:

- JSON 
- Form `date_select` boxes, i.e `date_of_birth: { "date(1i)" => "2012", "date(2i)" => "11", "date(3i)" => "28" }`
- Form text input field, depending on the current user's locale i.e `dd.mm.yyyy` or `mm/dd/yyyy`

The first two scenarios can be handled by a stateless InputCompiler, but the last one needs the current user's locale as state variable.

So it would be nice if we could re-configure the schema on-thy-fly by passing the options:

JSON:
`PersonSchema.with(input_processor: :json).call(json_params)`

Custom stateful input compiler:
```
input_processor = CustomFormInputCompiler.new(date_format: current_user.date_format)
PersonSchema.with(input_processor: inputProcessor).call(json_params)
```

solnic:
Gonna close this one since we decided to remove input processor compilers in 1.0.0 in favor of explicit type specs.

fran-worley:
@solnic are you keeping the branch around for any particular reason?

solnic:
Nope

PR #106, opened 2016-04-13
Fix bug when validating custom types

coop:
I was interested in looking at https://github.com/dry-rb/dry-validation/issues/103 - I've managed to get a passing spec (without breaking the build) but I can't help feeling that that this isn't the _actual_ solution. I'd love to seeing another failing spec.


solnic:
I pushed the actual solution :)

coop:
@solnic I _had_ that as my original solution but was able to get the tests to pass using `DEFAULT_TYPE_NODE.first` - I'm guessing the changes you made to the test exposed the need for `type.rule.to_ast`.

solnic:
@coop yes they did because otherwise everything would be treated as a string so no coercions would kick in

PR #108, opened 2016-04-14
Remove attr

coop:
I've pushed the original to https://github.com/dry-rb/dry-validation/tree/remove-attr so someone else can come along and implement at a later stage.


spec/unit/hint_compiler_spec.rb
```diff
@@ -23,14 +23,14 @@
...
+          [:val, [:predicate, [:key?, [:height]]]],
           [
             :or, [
               [:attr, [:height, [:predicate, [:none?, []]]]],
               [
                 :and, [
-                  [:attr, [:height, [:predicate, [:int?, []]]]],
-                  [:attr, [:height, [:predicate, [:gt?, [180]]]]]
+                  [:key, [:height, [:predicate, [:int?, []]]]],
+                  [:key, [:height, [:predicate, [:gt?, [180]]]]]
```
coop: The tests passed without making this change but I believe that is because dry-logic supports `attr?` as a predicate - I updated the tests to reflect the supported use cases.

Issue #114, opened 2016-04-23
Custom predicate error messages


tak1n:
Currently for custom predicate messages only `%{value}` is available.
For custom predicates which take 2 arguments (value and maybe a numeric limit etc) there is no way to output the limit which was used for that predicate.

tak1n:
What I could come up with now: https://github.com/tak1n/dry-validation/commit/f061a13595978afc76c88aef0e0c1b2ad320ef83

Not sure if there is a better solution to that, would also be cool to specify the argument by the same name  which it got in the custom predicate.
My knowledge about dry-v is still quite limited so not sure if we can get it somehow via: http://ruby-doc.org/core-2.2.0/Method.html#method-i-parameters

solnic:
Please do open a PR and we can discuss the approach there.

jeremyf:
@tak1n Can this issue be closed?

fran-worley:
@jeremyf not yet. The PR was accidentally merged and then reversed. We are looking at ways of making the param names available to avoid names like arg0. Please see the open PR in dry-logic and issue #165 for more details.

fran-worley:
This has now been addressed and a passing spec added via https://github.com/dry-rb/dry-validation/commit/1afd86eb0b8ed2f6c59ad2f580ec86681c39321e

PR #117, opened 2016-04-28
Custom predicate params in error messages

tak1n:
Currently there is no way to access all parameters of a custom predicate in the error message.
This pr just makes every parameter of a custom predicate method available (value is already available).

For now this means all args are accessible by `arg0`, `arg1` and so on.

Relates to #114

### DIFF
--- a/lib/dry/validation/error_compiler/input.rb
+++ b/lib/dry/validation/error_compiler/input.rb
@@ -132,6 +132,9 @@ module Dry
         meth = :"options_for_#{predicate}"
 
         defaults = { name: rule, rule: rule, value: input }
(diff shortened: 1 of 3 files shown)


solnic:
Can you provide a test for this behavior?

tak1n:
@solnic sorry this got completely stale.
I will try to add a unit spec for `ErrorCompiler::Input#options_for`

solnic:
No worries. It'd be better to have an integration spec showing the whole thing. Compiler stuff is too lowel / unstable to cover with unit tests for now.

tak1n:
Okay :+1:

Issue #92, opened 2016-04-04
Simplify `rule` syntax


solnic:
It should be possible to achieve this:

``` ruby
rule(foo: [:bar, :baz]) do
  bar.true?.then(baz.filled?)
end

# as an equivalent of:

rule(foo: [:bar, :baz]) do |bar, baz|
  bar.true?.then(baz.filled?)
end
```

daniels:
The proposed syntax works for the example, but for cases where the dependencies aren't literal symbol keys I'm not sure it will make sense:

```ruby
rule(foo: [[:bar, 0], :baz]) do
  ???.true?.then(baz.filled?)
end

rule(foo: [[:bar, :sub, :value], :baz]) do
  ???.true?.then(baz.filled?)
end

bar = :"prefix_#{var}"
rule(foo: [bar, :baz]) do
  ???.true?.then(baz.filled?)
end
```
What would an implicit name be in these examples?

Maybe we could use a hash instead:

```ruby
rule(foo: {value: [:bar, 0], baz: :baz}) do
  value.true?.then(baz.filled?)
end

rule(foo: {value: [:bar, :sub, :value], baz: :baz}) do
  value.true?.then(baz.filled?)
end

bar = :"prefix_#{var}"
rule(foo: {bar: bar, baz: :baz}) do
  bar.true?.then(baz.filled?)
end
```

But I don't think this is much of an improvement over being explicit about the names through block arguments:

```ruby
rule(foo: [[:bar, 0], :baz]) do |value, baz|
  value.true?.then(baz.filled?)
end

rule(foo: [[:bar, :sub, :value], :baz]) do |value, baz|
  value.true?.then(baz.filled?)
end

bar = :"prefix_#{var}"
rule(foo: [bar, :baz]) do |bar, baz|
  bar.true?.then(baz.filled?)
end
```

I guess we could make the simple case work and keep the current syntax for other cases?

daniels:
If we changed the interface to:
```ruby
rule(id, *simple_deps, **nested_deps)
```
Then we could do:
```ruby
rule(:foo, :baz, value: [:bar, 0]) do
  value.true?.then(baz.filled?)
end

rule(:foo, :baz, value: [:bar, :sub, :value]) do
  value.true?.then(baz.filled?)
end

bar = :"prefix_#{var}"
rule(:foo, :baz, bar: bar) do
  bar.true?.then(baz.filled?)
end
```

PR #120, opened 2016-04-30
Make rule work without block params

daniels:
This is a stab at fixing #92. I guess it could be seen as an ugly hack though, adding methods to the value instance. 

I moved as much of the code as I could out of the conditionals. Clearer to me, but maybe I should have kept separate instance_exec invocations depending on if there are any args?

The spec part is copy and paste from the spec for the old syntax. Please let me know if I should do that through shared examples instead, or in a totally different way/place.

### DIFF
--- a/lib/dry/validation/schema/value.rb
+++ b/lib/dry/validation/schema/value.rb
@@ -71,13 +71,25 @@ module Dry
 
         def rule(id = nil, **options, &block)
           if id
-            val = Value[id]
-            res = val.instance_exec(&block)
+            deps = []
           else
             id, deps = options.to_a.first
-            val = Value[id]
-            res = val.instance_exec(*deps.map { |name| val.value(name) }, &block)
           end
+
+          val = Value[id]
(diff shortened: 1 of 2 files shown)


solnic:
It's not a hack, it's a way to implement it :) This is a highly-dynamic DSL so we're gonna have code like that.

> The spec part is copy and paste from the spec for the old syntax. Please let me know if I should do that through shared examples instead, or in a totally different way/place.

I actually think we should remove support for passing args to the block, our goal is to make the DSL as concise as possible and this change reduces duplication.

solnic:
I guess we could deprecate passing args and warn that it's no longer needed and remove completely for 1.0

daniels:
Started to convert the other specs and uses in lib to use blocks without params and realized that this falls apart when the deps aren't simply key-symbols. 😢

I'll try to find the time to make another attempt later today.

solnic:
This API was removed eventually.

PR #124, opened 2016-05-02
Use included_in? and excluded_from?

jodosha:
Use `included_in?` instead of `inclusion?` and `excluded_from?` instead of `exclusion?`.
These predicates are deprecated by `dry-logic` See https://github.com/dry-rb/dry-logic/pull/9.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,8 @@ source 'https://rubygems.org'
 
 gemspec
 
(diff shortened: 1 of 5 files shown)


AMHOL:
Thanks @jodosha

jodosha:
@AMHOL Thanks for merging! 😄

Issue #116, opened 2016-04-27
Can't specify custom error message for `key?` predicate with specific attribute


iBublik:
Having following errors file:
```yaml
en:
  errors:
    rules:
      foo:
        key?: 'Foo is missing'
        filled?: 'Foo must be filled'
```
And following schema:
```ruby
FooSchema = Dry::Validation.Schema do
  required(:foo).filled
end
```
Error message for `filled?` predicate works as expected:
```ruby
FooSchema.call(foo: '').messages # => { foo: ['Foo must be filled'] }
```
But there's some problem with message for `key?` predicate:
```ruby
FooSchema.call({}).messages # => Dry::Validation::MissingMessageError: message for key? was not found
```

solnic:
I just realized why this happens. `key?` *is not* applied to the value, it is applied to the whole hash, so there's no rule for `foo` value, there's rule for the whole hash that it must have a key called `foo`.

iBublik:
So there's no way to do what I want?

solnic:
Not at the moment. I'll mark this as a feature request.

solnic:
We should probably add a pattern-matching algo to error compilation so that we can have messages defined like that:

``` yaml
en:
  errors:
     key?:
       value:
         foo: "FOO MUST BE PRESENT!!111"
```

ttdonovan:
I've submitted a pull request for this issue and am open to any additional feedback.

PR #142, opened 2016-05-06
add ability to specify custom error message for 'key?'

ttdonovan:
I've updated the `Validation::Messages::Abstract` class `lookup_path` to allow the ability to specify a custom error message for the `key?` predicate.

I've added the spec to `error_compiler_spec.rb` but am uncertain if that is the right spec that conveys the desired behavior.

This should address issue #116.

Any additional feedback is welcomed.


solnic:
oh noez I missed this one! could you rebase on top of master so that it merges cleanly?

ttdonovan:
@solnic I've gone ahead and rebased - this should now merge cleanly.

solnic:
Thank you very much :)

PR #158, opened 2016-05-11
enable coercion on form.true and form.false

### LINKED ISSUE #157: New macro: accepted
``` ruby
require 'bundler/setup'
require 'dry-validation'

schema = Dry::Validation.Schema do
  required(:terms).accepted
end

result = schema.(terms: '')
puts result.success? # => false
puts result.messages # => {:terms=>["must be accepted"]}
```

The macro should expand to `bool? & true?`.

---

I've noticed that a similar macro `confirmed` isn't implemented directly on `Dry::Validation::Schema::Rule`, but on `Dry::Validation::Schema::Value`, that translates to the following syntax:

``` ruby
schema = Dry::Validation.Schema do
  required(:password).filled(min_size?: 12).confirmation
end
```

Are macros that chains `required` allowed?

fran-worley:
Also add specs

fixes #157 

we no longer need an accepted macro instead you can just call `required(:terms).value(:true?)`

### DIFF
--- a/lib/dry/validation/input_processor_compiler/form.rb
+++ b/lib/dry/validation/input_processor_compiler/form.rb
@@ -5,6 +5,8 @@ module Dry
         default: 'string',
         none?: 'form.nil',
         bool?: 'form.bool',
+        true?: 'form.true',
+        false?: 'form.false',
         str?: 'string',
         int?: 'form.int',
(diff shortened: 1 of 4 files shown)


solnic:
Stellar job :)

PR #172, opened 2016-05-28
`each` not applying `array?` before hand when used with an `input_processor`

### LINKED ISSUE #170: predicate `array?` and `each` throwing an error
When using:
```ruby
required(:skills).filled(:array?)
```

It throws that error:
```
/home/mrbongiolo/.gem/ruby/2.3.0/gems/dry-types-0.7.2/lib/dry/types/compiler.rb:15:in `visit': undefined method `[]' for nil:NilClass (NoMethodError)
```

Also, when using `each` like this:
```ruby
required(:skills).each(:str?)
```

And calling the schema with: `{ skills: "some string" }` it throws this error:
```
 undefined method `map' for "some string":String
```

And I can't use `:array?` before `each` also because of the error above...

This is using the `master` branch.

mrbongiolo:
Added some failing spec for the case when `each` is used in a Schema with an `input_processor`.

Related to #170

### DIFF
--- a/spec/integration/schema/predicates/array_spec.rb
+++ b/spec/integration/schema/predicates/array_spec.rb
@@ -190,6 +190,58 @@ RSpec.describe 'Predicates: Array' do
           expect(result).to be_failing 0 => ['must be an integer']
         end
       end
+
+      context "when using an input_processor" do
+        subject(:schema) do
+          Dry::Validation.Schema do
+            configure do
+              config.input_processor = :sanitizer
+            end
+
+            required(:foo).each(:int?)
+          end
+        end
+
+        context 'with missing input' do
+          let(:input) { {} }
+
+          it 'is not successful' do
+            expect(result).to be_failing ['is missing']
+          end
+        end
+
+        context 'with nil input' do
+          let(:input) { { foo: nil } }
+
+          it 'is not successful' do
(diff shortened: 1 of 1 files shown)


mrbongiolo:
```ruby
def array_node(members)
  [:type, ['array', members]]
end
```

This line is part of the error, if changed to `'json.array'` or `'form.array'` it will apply the rule correctly.

solnic:
i'll be fixing issues with `each`, there are many things I should improve there, although this one looks suspicious :)

solnic:
btw this is probably related to a major issue we have with how hash schemas from dry-types work, see dry-rb/dry-types#87

mrbongiolo:
Ok, let me know if you need some help with anything, I still don't understand how everything is working under the hood, but I can try :)

solnic:
Conceptually, it's very simple - dry-v builds a hash schema type from dry-types based on type expectations in validation schemas. This type constructor is used prior applying validation rules. We started discovering various use cases where current logic is not what we want. So, when it comes to the details, it's pretty complex. The hash schema from dry-types must process all input, it shouldn't choke on any type errors, and it should return original values when potential coercion failed. And so I gotta improve it, by closing dry-rb/dry-types#87 we're gonna probably remove many known bugs in dry-v so it's my high priority task.

PR #177, opened 2016-06-06
Updates to specs following commit: #6a5b6ca

fran-worley:
Stop skipping specs
update expectations to match current behaviour
remove form/predicates/key_spec as this spec shouldn’t be used in forms.


fran-worley:
@solnic what do you think? I will stop committing to this PR now 😄

Issue #160, opened 2016-05-12
Make it possible to specify a predicate that has no args


solnic:
We need a way for composing predicates even when they don't require any args, so ie this can work:

``` ruby
Dry::Validation.Form do
  configure do
    def activation_code_required?
       # some logic which figures this out and relies on injected deps
       # so it doesn't need any input args
    end
  end

  required(:activation_code).maybe

  rule(new_user: [:activation_code]) do |activation_code|
    activation_code_required?.then(activation_code.filled?)
  end
end
```

/cc @apotonick

fran-worley:
@solnic I think (if I understood @apotonick correctly) we need to go a step further and be able to define rules that don't rely on a key at all. In your example above you are still hooking the rule to a key (:activation_code).

solnic:
`activation_code_required?` doesn't rely on any value, that's what Nick is asking for

fran-worley:
Unless I'm missing something, it does require the key

solnic:
high level rule that doesn't rely on any values makes no sense, the trick is to be able to compose predicates that rely on external information, ie something that queries a db etc.

fran-worley:
@solnic The implementation as you described it would be really useful I have quite a few examples where I have to pass in the value only to ignore it as I don't need it. So I'm saying :+1:

solnic:
This shouldn't be that hard (famous last words) as predicates have -1 arity so we can easily apply them w/o any input, it's a matter of looking what crashes right now and remove/tweak the code which assumes that we always pass in values when applying predicates. I *think* a `Rule::Check` from dry-logic w/o any value evaluator would do the trick, as it would simply apply its predicate with empty args.

fran-worley:
@solnic I'm working on this as we speak. I'll PR when I have something!

apotonick:
Thanks for mentioning me here, this is the only thing we need to release Reform 2.2 (which will define dry-v as default).

fran-worley:
@apotonick is the implementation described here correct? We are still hooking validation to a key, it's just that we will make the input optional.

apotonick:
Oh, you're right @fran-worley ! We need field-less "rules"!

fran-worley:
@apotonick can you provide a use case as I am struggling to come up with one.

apotonick:
In Reform-speak: :stuck_out_tongue_winking_eye: 

```ruby
collection :items do
  # ..
end

validate do
  items.size <= 5 
end
```

[...]

PR #180, opened 2016-06-07
[WIP] Use PredicateRegistry for setting up schemas

### LINKED ISSUE #165: Change the DSL so we have access to the predicate registry
See https://github.com/dry-rb/dry-logic/pull/13 for the background to this issue.

To quote @solnic 

> I planned to make an improvement in the dsl part of dry-v so that when we're generating rule ast we also already have access to predicate registry so that we can raise errors when invalid predicates are being defined. When you have a predicate registry, you also have access to arg names, so we could build an ast including names.
>
>Caveat: a schema object (not its class) has predicates too, and we invoke the DSL before a schema instance is created, so it's a chicken-egg issue. We could look into create a predicate registry for the DSL by merging configured Predicates module (class-level access, so we have it before a schema object is created) and somehow extract predicate methods from a schema.
>
>One solution would be to actually change how accessing predicates on a schema object works. Currently we have Schema#[] which can return an existing predicate if it's in the configured class-level registry, or see if a schema implements that predicate, then it returns a Dry::Logic::Predicate instance and pass in method object as the predicate function. Now, the problem here is that we don't have these predicates when DSL is invoked, so now I'm thinking about changing it so that we actually create unbound predicates using method_added hook and bind them when creating a schema instance and passing that registry, with all schema predicates bound to its instance, to the rule compiler.
>
>There are actually nice benefits of doing so, first of all we would have an explicit representation of schema-bound predicates, and we could easily avoid schema rule recompilation when using Schema#with and only recompile rules that have predicates that must be bound to a schema.
>
>So yeah, I feel like this is the way to go with this. It's gonna be a really nice improvement.

solnic:
closes #165

### DIFF
--- a/lib/dry/validation/predicate_registry.rb
+++ b/lib/dry/validation/predicate_registry.rb
@@ -60,7 +60,7 @@ module Dry
         if key?(name)
           predicate = self[name]
 
-          if predicate.arity != args.size + 1
+          if ![0, args.size + 1].include?(predicate.arity)
             raise_invalid_arity_error(name)
           end
         else
--- a/lib/dry/validation/predicate_registry.rb
+++ b/lib/dry/validation/predicate_registry.rb
@@ -14,13 +14,17 @@ module Dry
       class Unbound < PredicateRegistry
         def bind(schema)
           bound_predicates = predicates.each_with_object({}) do |(n, p), res|
-            res[n] = Dry::Logic::Predicate.new(n, &p.bind(schema))
+            res[n] = p.bind(schema)
           end
           Bound.new(external, bound_predicates)
         end
 
         def update(other)
-          predicates.update(other)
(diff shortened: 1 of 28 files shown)


lib/dry/validation/predicate_registry.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      def key?(name)
+        predicates.key?(name) || external.key?(name)
+      end
+
+      def ensure_valid_predicate(name, args)
+        if key?(name)
+          predicate = self[name]
+
+          if predicate.arity != args.size + 1
```
fran-worley: This will cause problems when we add support for predicates w/o args. We should probably check that arity !=0 before checking that there is space for the input arg.

fran-worley:
Will also fix #160

PR #185, opened 2016-06-15
transform args without options for

fran-worley:
Step 1 of cleaning up errors.

thoughts @solnic ??


config/errors.yml
```diff
@@ -71,10 +71,10 @@ en:
     size?:
       arg:
         default: "size must be %{size}"
-        range: "size must be within %{left} - %{right}"
+        range: "size must be within %{size_left} - %{size_right}"
```
solnic: what's the reason for renaming these vars?

fran-worley: Because we support custom predicates & blindly convert a range into 2 tokens for left and right if there was an existing arg called left or right it would get overriden. I figured that this was more explicit. The arg `size` transforms to  `size_left` and `size_right`.

PR #189, opened 2016-06-17
Separate type specs / coercions from predicates

### LINKED ISSUE #181: Separate type specs from predicates
Applying basic type-checking predicates should be enforced by the DSL, otherwise we may have schemas with pretty much broken rules leading to issues like #145 or #176 

To improve the situation we should have a new API that's dedicated for defining a type expectation and everything else will be built on top of that. My proposition:

``` ruby
Dry::Validation.Form do
  # => name must be present and it's a non-empty string (where :str type is default in Form)
  required(:name).filled

  # => age must be present and it's either an int or a nil
  required(:age, [:nil, :int]).maybe

  # => tags must be present and it's either an empty array or a 2-element
  #    array with string values
  required(:tags, :array) { empty? | size?(2) & each(:str?) }

  # => labels must be present and it's a non-empty array or string
  required(:labels, [:array, :str]).filled
end
```

closes #145 
closes #174 
closes #176 
closes #182 
closes #188

solnic:
This extends existing `required` and `optional` interfaces to support explicit type specs that are used for input coercion which is applied prior validation. This is a pretty big change as we're separating coercion from validation in the DSL too.

Here's how it works:

``` ruby
Dry::Validation.Form do
   # enable explicit type-specs
   configure { config.type_specs = true }

   required(:name, :string)
   required(:age, [:nil, :int])

   # `schema` implies :hash type spec
   required(:address).schema do
     required(:street, :string)
     required(:city, :string)
     required(:zipcode, :string)
   end

   # `each` implies `:array` type spec
   required(:phone_numbers).each do
     schema do
       required(:prefix, :string)
       required(:num, :string)
     end
   end
end
```

Notice that this is **a pure sanitization/coercion schema** without any rules.

Scope of this PR:

- [x] Extend `required` to accept type specs
- [x] Extend `optional` to accept type specs
- [x] Add support for type specs in nested schemas
- [x] Add support for either symbol-based type specs or dry-types objects
- [x] Make sure input-processor is applied only once (which will also fix jruby on CI)
- [x] Clean-up :)

As a second step, which will be done in another PR, I'll add a way to infer rules *from* type specs, which is exactly the opposite of what our complex input-processor compilers are doing. I also suspect some further polishing of the interface directly in master prior 0.8.0 release.

A really cool side-effect of this improvement is: defining a schema is ~80-100 x times faster no, no kidding:

```
Schema defined in 0.075425
Form defined in 5.43349
Form (with type specs) defined in 0.077061
------------------------------------------------------------------------------
              Schema     1.000  i/100ms
                Form     1.000  i/100ms
Form with type specs     1.000  i/100ms
              Schema     15.248  (±13.1%) i/s -     75.000  in   5.045985s
                Form      0.174  (± 0.0%) i/s -      1.000  in   5.749034s
Form with type specs     14.810  (±20.3%) i/s -     71.000  in   5.000067s

Comparison:
              Schema:       15.2 i/s
Form with type specs:       14.8 i/s - same-ish: difference falls within error
                Form:        0.2 i/s - 87.66x slower
```

closes #181

### DIFF
--- a/lib/dry/validation.rb
+++ b/lib/dry/validation.rb
@@ -7,6 +7,8 @@ require 'dry/validation/schema/form'
 require 'dry/validation/schema/json'
 
 module Dry
+  Types.register('form.string', Dry::Types['string'])
+
   module Validation
     MissingMessageError = Class.new(StandardError)
     InvalidSchemaError = Class.new(StandardError)
@@ -33,10 +35,7 @@ module Dry
         config.rules = config.rules + (options.fetch(:rules, []) + dsl.rules)
         config.checks = config.checks + dsl.checks
         config.path = dsl.path
-
-        if config.type_specs
-          config.type_map = type_map(dsl.type_map, config.input_processor)
-        end
+        config.type_map = klass.build_type_map(dsl.type_map) if klass.config.type_specs
       end
 
       if options[:build] == false
@@ -52,29 +51,6 @@ module Dry
 
     def self.JSON(options = {}, &block)
       Validation.Schema(Schema::JSON, options, &block)
-    end
-
-    def self.type_map(type_map, category)
-      type_map.each_with_object({}) do |(name, spec), result|
-        result[name] =
-          case spec
-          when Hash
-            Types["#{category}.hash"].symbolized(spec)
-          when Array
-            if spec.size == 1 && spec[0].is_a?(Hash)
-              member = Types["#{category}.hash"].symbolized(type_map(spec[0], category))
-              Types["#{category}.array"].member(member)
-            else
-              spec
-                .map { |id| id.is_a?(Symbol) ? Types["#{category}.#{id}"] : id }
-                .reduce(:|)
-            end
-          when Symbol
-            Types["#{category}.#{spec}"]
-          else
-            spec
-          end
-      end
     end
   end
 end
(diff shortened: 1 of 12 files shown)


solnic:
No more WIP. WDYT? /cc @AMHOL @fran-worley @jodosha @flash-gordon @mrbongiolo

fran-worley:
@solnic looks awesome. Should we update all the specs to use the new definition syntax?

solnic:
@fran-worley we should update all specs but first I gotta add type=>predicate inference so that we can simply write `required(:foo, :str).filled` vs `required(:foo, :str).filled(:str?)` or `required(:bar, [:nil, :str])` vs `required(:bar, [:nil, :str]).maybe(:str?)`

fran-worley:
@solnic happy to help update the specs. I think most of the pendings could be made live with this change too :)

solnic:
@fran-worley that would be great, thanks!

fran-worley:
You'll have to sort the jRuby issues... Definitely not my bag!

solnic:
@fran-worley yeah I'll deal with that over the weekend, ran out of time today already

timriley:
Reading through this again, this jumped out at me:

> Notice that this is a pure sanitization/coercion schema without any rules.

This is the kind of thing I could us implementing closer to the routing/request handling layer of an app, so that the data being passed from it to an app's operations is coerced into proper types. However, if that operation then wanted to apply its own validation scheme with real business rules included, would it need to re-state all those type specs?

solnic:
@timriley I keep thinking about this but I didn't manage to figure out how this should work in a way where you don't have to, well, repeat yourself too much :) We should explore this idea more prior 1.0.

timriley:
Cool, sounds good 👍🏼

PR #203, opened 2016-07-04
Reduce method calls by caching as ivar lazily

splattael:
dbf9534 made benchmarks fast but tests slow so it has been
reverted in 944558f

This commit make benchmarks _and_ tests fast(er) again by
lazily caching the cache in a ivar.

/cc @solnic


splattael:
> @solnic: pls just merge it in once travis is green

Travis is :green_heart:

PR #219, opened 2016-07-19
Updated errors.yml

mrbongiolo:
Updated the keys included_in? and excluded_from? to show the correct message when used with a range.


mrbongiolo:
Done!

solnic:
@mrbongiolo thanks! I'll include that in 0.9.3

Issue #216, opened 2016-07-19
I18n::InvalidLocale: :en is not a valid locale


mrbongiolo:
This happened when upgrading from `0.8.0` to `0.9.x`.

In my rails app I have the I18n locales defined like that, on application.rb file:
```ruby
    I18n.config.enforce_available_locales = true
    config.i18n.available_locales = ["pt-BR"]
    config.i18n.default_locale = "pt-BR"
```

The current version of dry-v ignores the `I18n.default_locale` option in favor of it's `:locale` option (that can be set when calling `results.messages`), since I haven't used the `:locale` option there then it defaulted to `:en`.

Current solution: call `results.messages(locale: 'pt-BR')`

Suggestion: add a config option to set the "default" locale, similar to the way we can set `messages = :i18n`.

agustin:
Hi! I'm also struggling with a similar situation in which `I18n` is configured and without monkey patching **[Reform](https://github.com/apotonick/reform)** ([Issue #391](https://github.com/apotonick/reform/issues/391)) I'm not able to pass the `locale:` option to the `messages(...)` but working on it I encounter that the there is something unclear on why is implemented the way it is, and I believe that I'm missing something or have a lack of background, that's why I'm asking in here.

I wonder the following, why someone will define the use of `:i18n` for the messages:
```ruby
Dry::Validation.Schema do
  configure do
    config.messages = :i18n
  end
  #[...]
end
```
and then need to pass it as an argument to the `messages(...)`, shouldn't make more sense in this case to follow the configurations on the `I18n` Module? like the `I18n.locale` or `I18n.default_locale`?
I worked on a commit for this approach, but thought that maybe there is a reason that I'm not aware of that justify this behavior. (Happy to push the commit and discuss it over there also)

Anyone knows? maybe @solnic can give us some better insight on this 🔦 😃 
Thanks!

solnic:
This is a simple fix probably. There's a mistake in the logic as message compiler defaults to `:en` [right here](https://github.com/dry-rb/dry-validation/blob/master/lib/dry/validation/message_compiler.rb#L14) so the only thing we need to do is to add `Messages::Abstract#default_locale` and implement it in `Messages::YAML` and `Messages::I18n` and tweak that line to do `@locale = options.fetch(:locale, messages.default_locale)`.

cavi21:
That sounds perfect @solnic! do you want me to create a PR for that? or you rather do it by yourself?
[...]

PR #235, opened 2016-08-15
Use I18n.locale or I18n.default_locale when the config specify :18n for the messages

agustin:
Based on the conversation in the Issue **#216** the situation was that the default_locale was fixed on `:en` and thus also when you were setting the `config.messages = :i18n` you were obliged to pass the `locale` as a parameter to the messages.

@solnic let me know if this is what you was expecting or you had a different idea?

Thanks! 👍

### DIFF
--- a/lib/dry/validation/messages/i18n.rb
+++ b/lib/dry/validation/messages/i18n.rb
@@ -18,7 +18,8 @@ module Dry
       end
 
       def key?(key, options)
-        ::I18n.exists?(key, options.fetch(:locale, default_locale))
+        ::I18n.exists?(key, options.fetch(:locale, default_locale)) ||
+        ::I18n.exists?(key, I18n.default_locale)
       end
 
       def merge(path)
--- a/lib/dry/validation/messages/i18n.rb
+++ b/lib/dry/validation/messages/i18n.rb
@@ -18,12 +18,16 @@ module Dry
       end
 
       def key?(key, options)
-        ::I18n.exists?(key, options.fetch(:locale, I18n.default_locale))
+        ::I18n.exists?(key, options.fetch(:locale, default_locale))
       end
 
       def merge(path)
         ::I18n.load_path << path
         self
+      end
+
+      def default_locale
+        I18n.locale || I18n.default_locale || super
       end
     end
   end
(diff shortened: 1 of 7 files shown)


spec/fixtures/locales/pl.yml
```diff
@@ -5,6 +5,10 @@ pl:
       arg:
         default: "wielkość musi być równa %{size}"
         range: "wielkość musi być między %{size_left} a %{size_right}"
+      value:
+        string:
+          arg:
+            default: "wielkość musi być równa %{size}"
```
cavi21: Not sure if this is correct :/!

lib/dry/validation/messages/i18n.rb
```diff
@@ -18,7 +18,8 @@ def get(key, options = {})
       end
 
       def key?(key, options)
-        ::I18n.exists?(key, options.fetch(:locale, default_locale))
+        ::I18n.exists?(key, options.fetch(:locale, default_locale)) ||
+        ::I18n.exists?(key, I18n.default_locale)
```
solnic: since `#default_locale` returns `I18n.default_locale` - why do we need this?

solnic:
I only had one question, please let me know and I'll merge this in and push a release. Thanks for helping with this.

lib/dry/validation/messages/i18n.rb
cavi21: Hi! the reason behind this is because could be the case that someone has `I18n` configured to use [fallbacks](https://github.com/svenfuchs/i18n/wiki/Fallbacks) but the `exists?`call return `false` on the `key?` and so never fetch afterwards. 

With this extra check, the `get(key, options = {})` will call `translate` on the I18n module and it will handle the fallback properly.

Also I know that is not as accurate as it could be, because we are not necessary checking on all the fallback chain, but if someone has I18n fallback configured with a default one probably this will have at least all the keys, I'm assuming this though 😕 

I just push a spec that shows the reason behind this

cavi21:
I just push a rebased commit with that additional OR and the corresponding spec. Also try to explain a little bit better the reason behind it.

Let me know what you think? and happy to help and discuss this if you are not convince about it!
Thanks!

solnic:
Oh I see, yeah this makes sense. Thanks again! Merging/releasing this.

PR #237, opened 2016-08-17
Add methods to clear message cache

yoshitsugu:
I want some way to clear cache for messages.
This is because I'm now creating some application which error messages changing dynamically.
I don't know whether this is the common use case, but if someone is needed, please add this kind of feature.
If dry-validation already has this feature, please tell me.
Thank you.

### DIFF
--- a/lib/dry/validation/messages/abstract.rb
+++ b/lib/dry/validation/messages/abstract.rb
@@ -42,6 +42,10 @@ module Dry
           @cache ||= Concurrent::Map.new { |h, k| h[k] = Concurrent::Map.new }
         end
 
+        def self.clear_cache!
+          @cache = nil
+        end
+
         attr_reader :config
 
         def initialize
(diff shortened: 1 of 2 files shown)


solnic:
I was planning to centralize message cache and make it configurable (as in you can opt-in/out) and add a method to clear it on demand. It's a bit more tricky than it seems to be though, so I gotta figure out how to do it in a nice way, I also want to use some external utility for that as we use caching in lots of gems and it's time to finally unify it.

For now I would recommend using whatever works for you through monkey-patches, and I'll come up with a final solution in version 0.10.0.

How does that sound?

yoshitsugu:
Thank you for you reply.
I got it. It sounds nice to me because I don't understand the library so I have no confidence about my change.
I will close this pull request for now and I'm very looking forward to version 0.10.0!

yoshitsugu:
@solnic Is there any progress for this problem?

solnic:
Not yet. I didn't manage to make it for 0.10.0. I'll come up with some solution before 1.0.0.

Issue #229, opened 2016-08-04
Support Dry::Validation::Result#to_either


saverio-kantox:
To properly chain validations with other more 'monadic' operations, I found the need of something like
```ruby
MySchema.(params).to_either
.fmap { |output| do_stuff(output) }
.bind { |whatever| else }
```

Do you think it makes sense to have a native `to_either`, plus delegated `bind` and `fmap` directly on the result? In that case what would be the desired content of the Right, the result itself or the output? And the Left would contain the result or just the errors?

solnic:
This sounds like a good idea. We could add an optional monad extension for `Result`. I suppose `Left` should include error messages, and `Right` should include result output. WDYT @flash-gordon @timriley ??

flash-gordon:
👍

timriley:
Yep, this seems good to me 👍

I wondered for a moment as to whether the `Left` side should also contain the output too, but I realise that in most situations, the _original_ input (i.e. before it was passed to the schema) should still be available to reuse in the error case anyway.

PR #238, opened 2016-08-18
Add support for .to_either on Result object

flash-gordon:
See #229

### DIFF
--- a/lib/dry/validation/result.rb
+++ b/lib/dry/validation/result.rb
@@ -1,9 +1,11 @@
+require 'dry/monads/either'
 require 'dry/validation/constants'
 
 module Dry
   module Validation
     class Result
       include Dry::Equalizer(:output, :errors)
+      include Dry::Monads::Either::Mixin
       include Enumerable
 
       attr_reader :output
@@ -67,6 +69,14 @@ module Dry
 
       def name
         Array(path).last
+      end
+
+      def to_either(options = EMPTY_HASH)
+        if success?
+          Right(output)
+        else
+          Left(messages(options))
+        end
       end
 
       private
--- a/lib/dry/validation/result.rb
+++ b/lib/dry/validation/result.rb
@@ -1,11 +1,9 @@
-require 'dry/monads/either'
 require 'dry/validation/constants'
 
 module Dry
   module Validation
     class Result
       include Dry::Equalizer(:output, :errors)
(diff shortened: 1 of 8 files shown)


timriley:
Looks good 👍 :)

solnic:
This implicitly assumes dry-monads are present, which is true because dry-types depends on it, but I wouldn't want to make it depend on dry-monads, so I'm wondering if we should enable feature-detection somehow and include it as an optional extension. We *need* this already for dry-struct too (I don't want dry-v to have a hard-dep on dry-struct either).

flash-gordon:
@solnic should it be manually-required opt-in behaviour or something automagically-detected one? ie sutin like `if defined? Dry::Monads`

solnic:
We should probably try to require and rescue LoadError

lib/dry/validation/result.rb
```diff
@@ -81,3 +81,11 @@ def result_ast
     end
   end
 end
+
+begin
```
flash-gordon: @solnic how about this? :)

solnic: could we add it to something reusable like `Dry::Validation.load_extensions(:monads, :struct)` or something like that? this way we'd reduce duplication and be able to easily have tests for it

flash-gordon: Then using `Dry::Validation.load_extensions` does not suppose using `rescue LoadError`, am I right?

flash-gordon: ah, I see, it will work differently for dry-struct and dry-monads. I mean as long as dry-t depends on dry-m, we won't have LoadError. However, dry-struct will be required in specs only, so we won't try to require it in dry-v, leaving that step for the user, right (by calling `Dry::Validation.load_extensions(:struct)`)?

solnic: dry-t shouldn't depend on dry-m so we need this feature there too...dry-core? :D

flash-gordon: @solnic would it be enough https://github.com/dry-rb/dry-core/pull/3?

solnic: Yes. I merged it in

flash-gordon:
@solnic have a look

solnic:
This is really awesome :)

PR #243, opened 2016-08-29
Add custom messages

kzielonka:
In my current project validation shouldn't just return messages but objects with fields like :error_code (for example `INVALID_EMAIL_ERROR` and :message (which is an normal message like `Email format is invalid.`).
I didn't know how to do it using :yaml or :i18n messages providers so I created this pull request.

I extended `Dry::Validation::Schema` to accept custom messages object.
The setting option `messages` can now be one of :yaml, :i18n or :custom.
If :custom option is set `custom_messages` should also be set as Object which implements #default_locale and #[](key, options) methods.
Method #[](key, options) should return objects which implement #% (for example Strings but not necessary) or is a subclass of `Dry::Validation::Messages::Abstract` (please check tests for more details how it can be sued).

Plese check is it ok and I hope that it will be merged into master ;)

### DIFF
--- a/spec/integration/custom_error_messages_spec.rb
+++ b/spec/integration/custom_error_messages_spec.rb
@@ -45,4 +45,96 @@ RSpec.describe Dry::Validation do
       include_context 'schema with customized messages'
     end
   end
+
+  context 'custom messages' do
+    context 'with invalid settings' do
+      it 'should return error when config.custom_messages is nil' do
+        expect do
+          Dry::Validation.Schema do
+            configure do
+              config.messages = :custom
+              config.custom_messages = nil
+            end
+
+            required(:email, &:filled?)
+          end
+        end.to raise_error RuntimeError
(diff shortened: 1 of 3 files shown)


spec/integration/custom_error_messages_spec.rb
```diff
@@ -45,4 +45,96 @@
...
+      end
+
+      context 'custom messages object returning custom message object' do
+        let(:message) do
+          obj = Object.new
+          obj.define_singleton_method(:%) { |_| self }
+          obj
+        end
+
+        before :each do
```
dsounded: just `before` ?

solnic:
This API was moved to dry-schema project. It's possible that its specialization will be re-added to dry-validation, for now this PR is outdated. We can get back to this when things are more settled in master, as the lib was rewritten on top of dry-schema.

PR #244, opened 2016-08-29
[WIP] Update and improve integration with dry-struct

### LINKED ISSUE #241: Update and improve integration with dry-struct
`Types::Struct` has been extracted to `Dry::Struct` in [dry-struct](https://github.com/dry-rb/dry-struct) gem and we gotta update dry-validation to use it. Also, this should be **optional**.
## TODO
- [x] extract struct-related functionality so it can be optionally included
- [ ] include struct-related support conditionally when `Dry::Struct` is present
- [ ] improve the integration so that complex structs with types like array with members or nested structs is also supported

cored:
Fix #241 

- Add `Dry::Struct`

### DIFF
--- a/dry-validation.gemspec
+++ b/dry-validation.gemspec
@@ -21,6 +21,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'dry-equalizer', '~> 0.2'
   spec.add_runtime_dependency 'dry-logic', '~> 0.3', '>= 0.3.0'
   spec.add_runtime_dependency 'dry-types', '~> 0.8', '>= 0.8.1'
+  spec.add_runtime_dependency 'dry-struct', '~> 0.0.1'
 
   spec.add_development_dependency 'bundler'
   spec.add_development_dependency 'rake'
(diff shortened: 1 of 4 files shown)


solnic:
I had to make this work with dry-struct already in another branch which is now merged into master, so you could just hard-reset to latest master and continue from there :)

cored:
Will do that, thanks

solnic:
Thanks to @flash-gordon we now support optional extensions via `Dry::Validation.register_extension(:foo) { something that must happen when it is enabled }`. You can see how this was used [here](https://github.com/dry-rb/dry-validation/commit/1bf33a893de4e088ec737c1ea1880e67ce20aa7b).

So, now I'd imagine we could refactor code a bit so that we load `:struct` extension as optional, via something like this:

``` ruby
Dry::Validation.register_extension(:struct) do
  require 'dry/validation/extensions/struct'
end
```

And in this file we'd just extend APIs to support inferring rules from structs. This *does need* some refactoring as struct integration has never been written as something optional, so now we gotta improve it :)

cored:
Interesting, I will check it out.

flash-gordon:
@solnic ah, I was about to write just the same :)

cored:
@solnic Closing this @flash-gordon already implemented it.

flash-gordon:
@cored well, not exactly 😄  We're still lack of support for arrays and nested structs. The part I implemented was the easiest tbh :)

PR #254, opened 2016-09-21
Run specs twice on CI (with and wo extensions)

flash-gordon:


Rakefile
```diff
@@ -6,7 +6,16 @@ $LOAD_PATH.unshift(File.join(File.dirname(__FILE__), 'lib'))
...
-task default: :spec
-
 desc 'Run all specs in spec directory'
-RSpec::Core::RakeTask.new(:spec)
+task :run_specs do
+  require 'rspec/core'
+
+  RSpec::Core::Runner.run(['spec/integration', 'spec/unit'])
+  RSpec.clear_examples
+  Dir[SPEC_ROOT.join('shared/**/*.rb')].each(&method(:load))
```
flash-gordon: @solnic pls, pay attention, `clear_examples` flushes shared examples, so we need to manually reload them to run specs for the second time

flash-gordon: oh, I pressed big green button and started a review apparently

spec/spec_helper.rb
```diff
@@ -40,4 +40,23 @@ module Types
   end
 
   config.include PredicatesIntegration
+
+  config.before do
```
flash-gordon: @solnic fyi, this is a copy & paste from dry-struct. We'll be able to remove container manipulations when https://github.com/dry-rb/dry-struct/issues/12 is resolved

flash-gordon: I updated this part

PR #255, opened 2016-09-21
Add dry-struct integration via an extension

flash-gordon:
This took me a dozen of minutes /cc @solnic @cored

### DIFF
--- a/lib/dry/validation/extensions/struct.rb
+++ b/lib/dry/validation/extensions/struct.rb
@@ -1,0 +1,32 @@
+require 'dry-struct'
+
+module Dry
+  module Validation
+    class Schema
+      module StructClassBuilder
+        def create_class(target, other = nil)
+          if other.is_a?(Class) && other < Dry::Struct
+            super do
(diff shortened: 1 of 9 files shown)


solnic:
😍

lib/dry/validation/extensions/struct.rb
```diff
@@ -0,0 +1,32 @@
+require 'dry-struct'
+
+module Dry
+  module Validation
+    class Schema
+      module StructClassBuilder
```
cored: @flash-gordon Why didn't you just created a class name `Struct < Dry::Struct` and at the test level just use that new class ?

flash-gordon: @cored could you please elaborate on this?

cored: @flash-gordon nvm, I need to understand the code better. My initial implementation had something like 
```ruby 
class Dry::Schema::Struct < Dry::Struct 
... 
end 
```

PR #281, opened 2016-12-01
require dry/validation/version

georgemillo:
`'dry/validation/version'` isn't required automatically, meaning if I want to check the version at runtime I need to `require` it myself. I'm not sure if this is a deliberate choice or an oversight, but the overwhelming majority of gems I've seen have a constant called `GemName::VERSION` (or `Version`) that's required by default, so I thought I'd make the suggestion here. :)


solnic:
I don't mind :) Thanks

Issue #292, opened 2016-12-27
Duplicate `require 'rspec/core` in Rakefile


ziwag:
I'm studying **dry-validation** code and I found a duplicate `require 'rspec/core'` in the Rakefile:

```ruby
require 'rspec/core'
require 'rspec/core/rake_task'

task :run_specs do
  require 'rspec/core'

  RSpec::Core::Runner.run(['spec/integration', 'spec/unit'])
  RSpec.clear_examples
  Dir[SPEC_ROOT.join('shared/**/*.rb')].each(&method(:load))

  Dry::Validation.load_extensions(:monads, :struct)
  RSpec::Core::Runner.run(['spec'])
end
```

I don't know if it's a normal behavior for older Ruby versions or a typo so I want point it.

flash-gordon:
thanks for spotting, this was fixed

PR #296, opened 2016-12-27
Removed duplicate 'rspec/core import in rake file

ggrumbley:
Hello, it seemed that there were two rspec/core imports in the project's Rakefile (per issue #292). I have removed the second one from inside the first task.


solnic:
Thanks!

PR #286, opened 2017-01-06
[WIP] messages(errors: :full) returns an array, not a hash

georgemillo:
[Original discussion here](https://discuss.dry-rb.org/t/dry-v-feature-suggestion-returning-an-array-not-a-hash-from-errors/166)

The cleanest way I could get this to work was for MessageCompiler to forget about `full` and just pass it straight into MessageSet#dump. But after writing this, I realised that there's a problem: MessageCompiler, not MessageSet, handles translations, and with my current approach there's no easy way to translate the *key*.

I can't think of a way to fix this that doesn't require major changes to the way MessageCompiler and MessageSet work, e.g. by making MessageSet rather than MessageCompiler handle the translations.

I'm opening the unfinished PR just to get the discussion going again, but I think my suggestion of  keeping the API the same (`messages(full: true)`) might not be feasible. Maybe we could go back to my very first suggestion from the discussion thread: to keep the existing API the same, but to add a second option:

```ruby
result.errors(full: true)
# => {age: ["must be present", "must be at least 18"], name: ["must be 20 characters or less"]}
result.errors(full: true)
# => {age: ["age must be present", "age must be at least 18"], name: ["name must be 20 characters or less"]}
result.errors(full: true, flat: true)
# => ["age must be present", "age must be at least 18", "name must be 20 characters or less"]
```

Although I can't think of a situation where you'd want `flat` to be true but `full` to be false, so maybe this makes more sense:

```ruby
result.errors(full: true)
# => {age: ["must be present", "must be at least 18"], name: ["must be 20 characters or less"]}
result.errors(full: true)
# => {age: ["age must be present", "age must be at least 18"], name: ["name must be 20 characters or less"]}
result.errors(flat: true)
# => ["age must be present", "age must be at least 18", "name must be 20 characters or less"]

# another idea; but I prefer the above one:
result.errors(full: :flat)
# => ["age must be present", "age must be at least 18", "name must be 20 characters or less"]
```

What do you think?

### DIFF
--- a/lib/dry/validation/message_compiler.rb
+++ b/lib/dry/validation/message_compiler.rb
@@ -19,10 +19,6 @@ module Dry
         @default_lookup_options = { locale: locale }
       end
 
-      def full?
-        @full
-      end
-
       def hints?
         @hints
       end
@@ -160,14 +156,7 @@ module Dry
       end
 
       def message_text(rule, template, tokens, opts)
-        text = template % tokens
-
-        if full?
-          rule_name = messages.rule(rule, opts) || rule
-          "#{rule_name} #{text}"
-        else
-          text
-        end
+        template % tokens
       end
 
       def message_tokens(args)
--- a/lib/dry/validation/message_set.rb
+++ b/lib/dry/validation/message_set.rb
@@ -27,8 +27,12 @@ module Dry
         initialize_placeholders!
       end
 
-      def dump
-        root? ? to_a : to_h
+      def dump(full: false)
+        if full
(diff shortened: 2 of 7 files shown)


georgemillo:
Just converted this issue to a PR (I didn't know you could do that!!! Although it is deprecated :( ), added some actual code, and updated my original comment. Please advise!

solnic:
Closing as now all this code is in dry-schema. We could add this feature there and then update related code here to support it too.

PR #318, opened 2017-03-06
Refactor test docstrings

dgollahon:
I found these while test-driving a `rubocop-rspec` cop that checks for redundant "it"s in test docstrings.

- Removes redundant 'it's at the beginning of test docstrings.
- Converts 'should raise error' and 'is raises error' to just
  'raises error' which is consistent with the other examples in
  the codebase.

I'm happy to make similar changes to other `dry-*` repos if that would be helpful/desired, just lmk!


solnic:
Thanks!

PR #323, opened 2017-04-04
Input types for `hash?` and `array?` fields in JSON schema

flexoid:
Fixed input types that are generated for `Validator.JSON` schemas
when fields with `hash?` and `array?` are defined without corresponding
nested schema:

```ruby
schema = Dry::Validation.JSON do
  required(:attributes).value(:hash?)
  required(:ids).value(:array?)
end

schema.input_processor
```

was result in:
```
#<Dry::Types::Safe
  type=#<Dry::Types::Hash::Symbolized primitive=Hash
    options={:member_types=>{
      :attributes=>#<Dry::Types::Safe type=#<Dry::Types::Definition primitive=String options={}> options={}>,
      :ids=>#<Dry::Types::Safe type=#<Dry::Types::Definition primitive=String options={}> options={}>}}>
  options={}>
```

Note `primitive=String` on both `:attributes` and `:ids` fields.

Now fixed to be `Hash` and `String` respectively.

=========

I found this issue when investigating huge performance loss with my API classes, and notices what this code executed many times on failure branch, and constantly doing "inspect" on quite large incoming data was slowing down the code.

https://github.com/dry-rb/dry-types/blob/v0.9.4/lib/dry/types/definition.rb#L69

```ruby
def try(input, &block)
  if valid?(input)
    success(input)
  else
    failure = failure(input, "#{input.inspect} must be an instance of #{primitive}")
  block ? yield(failure) : failure
  end
end
```

### DIFF
--- a/lib/dry/validation/input_processor_compiler/json.rb
+++ b/lib/dry/validation/input_processor_compiler/json.rb
@@ -11,7 +11,9 @@ module Dry
         decimal?: 'json.decimal',
         date?: 'json.date',
         date_time?: 'json.date_time',
-        time?: 'json.time'
+        time?: 'json.time',
+        hash?: 'json.hash',
+        array?: 'json.array'
       }.freeze
 
       CONST_MAP = {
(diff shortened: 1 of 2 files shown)


solnic:
This API is now gone, type-specs are always used in dry-schema which will be the schema backend in dry-validation 1.0.0.

PR #325, opened 2017-04-25
Fixes issue with wrong localized error messages

### LINKED ISSUE #322: Combining i18n with namespace for error messages
It seems that when I combine `i18n` with `namespace` configuration the message root locale goes back to the default `:en`.
Here is my `config/locales/errors.pl.yml` file:
```yaml
pl:
  errors:
    filled?: Pole nie może pozostać puste
    str?: Wartość musi być tekstem

    rules:
      note:
        rules:
          note:
            filled?: Alternatywny komunikat o pustym polu
```

Here is the code:
```ruby
      schema = Dry::Validation.Form do
        configure {
          config.namespace = :note
          config.messages = :i18n
        }
        required(:note).filled(:str?)
      end

      schema.call(params.to_unsafe_h).messages
```

In order to make it work properly I have to set locale explicitly as a messages `locale` argument:
```ruby
      schema = Dry::Validation.Form do
        configure {
          config.namespace = :note
          config.messages = :i18n
        }
        required(:note).filled(:str?)
      end

      schema.call(params.to_unsafe_h).messages(locale: I18n.default_locale)
```
I don't think it is supposed to work that way, but I may be mistaken.

I run this code inside Rails 5 app with `rails-i18n` gem.

kbredemeier:
Closes #322

### DIFF
--- a/spec/integration/localized_error_messages_spec.rb
+++ b/spec/integration/localized_error_messages_spec.rb
@@ -49,6 +49,24 @@ RSpec.describe Dry::Validation, 'with localized messages' do
           )
         end
       end
+
+      describe '#errors' do
+        context 'with different locale' do
+          before do
+            I18n.locale = :pl
+          end
+
+          after do
+            I18n.locale = :en
+          end
+
+          it 'contains the localized errors' do
+            expect(schema.(email: '').errors).to eql(
+              { email: ['Hej user! Dawaj ten email no!'] }
+            )
+          end
+        end
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/integration/localized_error_messages_spec.rb
```diff
@@ -49,6 +49,24 @@
...
+            I18n.locale = :pl
+          end
+
+          after do
+            I18n.locale = :en
+          end
+
+          it 'contains the localized errors' do
+            expect(schema.(email: '').errors).to eql(
+              { email: ['Hej user! Dawaj ten email no!'] }
```
solnic: 😆 ❤️

solnic:
Thank you!

PR #329, opened 2017-05-03
Bug when injecting rules into a schema

pabloh:
When injecting rules from to a new schema the original's rules are also modified


solnic:
Nice catch, thank you

PR #341, opened 2017-06-21
Update the compiler to work with latest version of dry-types

GustavoCaso:
Update compiler to work with the latest version of `dry-types`

### DIFF
--- a/lib/dry/validation/input_processor_compiler/json.rb
+++ b/lib/dry/validation/input_processor_compiler/json.rb
@@ -34,11 +34,11 @@ module Dry
       end
 
       def hash_node(schema)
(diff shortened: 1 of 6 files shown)


lib/dry/validation/input_processor_compiler/json.rb
```diff
@@ -34,11 +34,11 @@ def identifier
       end
 
       def hash_node(schema)
-        [:type, ['json.hash', [:symbolized, schema]]]
+        [:json_hash, [:symbolized, schema, {}]]
       end
 
       def array_node(members)
-        [:type, ['json.array', members]]
+        [:json_array, members, {}]
```
solnic: This should be `[:json_array, [members, {}]]`

PR #345, opened 2017-06-29
Fix `warning: extra states are no longer copied`

nestorsalceda:
Just a fix for removing the warning when rejecting extra states on a Hash

https://github.com/ruby/ruby/blob/v2_4_1/hash.c#L1335-L1337

Thanks!


flash-gordon:
Thank you :) It took a while but I'm finally working on the problem with warnings

Issue #353, opened 2017-07-25
Wrong translation key for rules


raykov:
This line https://github.com/dry-rb/dry-validation/blob/58699acf23d3c101e01662a8633197ca39b68696/lib/dry/validation/messages/abstract.rb#L58 generates translation keys like `%{locale}.rules.name` for `Dry ::Validation::Messages::I18n`. And can't find translation with this key. It looks like we should have `translation.yml` like this

```
en:
  "%{locale}":
    rules:
      name: Name
```

For `Dry ::Validation::Messages::YAML` it works fine, because here we have `key % { locale: :en}`
https://github.com/dry-rb/dry-validation/blob/58699acf23d3c101e01662a8633197ca39b68696/lib/dry/validation/messages/yaml.rb#L38

So, we need to fix this for `Messages::I18n`

Also we need to provide ability use namespaces for `rules`

regedarek:
hi

solnic:
This was fixed in dry-schema, thus it'll be fixed here in 1.0.0.

PR #354, opened 2017-07-25
[Issue #353] fix bug with wrong translation key in Messages::I18n#rule

raykov:
### What

Added method `rule_path` for `Messages::Abstract` and `Messages::YAML`. Because path for  `Messages::I18n` and `Messages::YAML` should be different.

For `Messages::I18n` we should have key like `"rules.#{name}"`. Because 
```
      def key?(key, options)
        ::I18n.exists?(key, options.fetch(:locale, default_locale)) ||
        ::I18n.exists?(key, I18n.default_locale)
      end
```
will not find key like `%{locale}.rules...`.

For `Messages::YAML` it should be `"%{locales}.rules.#{name}"`. Because
```
      def key?(key, options = {})
        data.key?(key % { locale: options.fetch(:locale, default_locale) })
      end
```

For `Messages::Namespaced` it will be build based on `messages` class.

### Related Issue

https://github.com/dry-rb/dry-validation/issues/353

### DIFF
--- a/lib/dry/validation/messages/abstract.rb
+++ b/lib/dry/validation/messages/abstract.rb
@@ -54,8 +54,12 @@ module Dry
           @hash ||= config.hash
         end
 
+        def rule_path(name)
+          "rules.#{name}"
+        end
+
         def rule(name, options = {})
-          path = "%{locale}.rules.#{name}"
+          path = rule_path(name)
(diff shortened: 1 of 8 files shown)


solnic:
This was fixed in dry-schema.

PR #359, opened 2017-08-22
[Issue #358] fix schema subclasses type_map inheritence

### LINKED ISSUE #358: Schema subclasses where parent has config.type_specs = true fail to inherit parent's type_map
below is a simple example to demonstrate

```
module Types
  include Dry::Types.module

  Uppercased = Types::String.constructor {|str| str.upcase }
end

RSpec.describe 'inherited schema with type_specs=true' do
  let(:single) do
    Dry::Validation.Form do
      configure do
        config.type_specs = true
      end

      required(:foo, Types::Uppercased).filled
      required(:bar, Types::Uppercased).filled
    end
  end

  let(:parent) do
    Dry::Validation.Form do
      configure do
        config.type_specs = true
      end

      required(:foo, Types::Uppercased).filled
    end
  end

  let(:child) do
    Dry::Validation.Schema(parent) do
      required(:bar, Types::Uppercased).filled
    end
  end

  it 'works correctly for single schema' do
    expect(single.call(foo: ?a, bar: ?b).output).to include(foo: ?A, bar: ?B)
  end

  it 'currently breaks in 0.11.0' do
    expect(child.call(foo: ?a, bar: ?b).output).to include(foo: ?A, bar: ?B)
  end
end
```
the final expectation is missing `foo: "A"`, the definition was not inherited by the child schema

mazniak:
This fixes #358 and also adds some hard InvalidSchema errors when attempting to subclass a schema defining an array type_spec with non-array type_spec definition and vice-versa

### DIFF
--- a/lib/dry/validation/schema/class_interface.rb
+++ b/lib/dry/validation/schema/class_interface.rb
@@ -61,7 +61,8 @@ module Dry
           cfg.rules = rules
           cfg.checks = cfg.checks + dsl.checks
           cfg.path = dsl.path
-          cfg.type_map = target.build_type_map(dsl.type_map) if cfg.type_specs
+          cfg.type_map = target.
+            build_inherited_type_map(dsl.type_map, config.type_map) if cfg.type_specs
         end
 
         target
--- a/lib/dry/validation/type_specs.rb
+++ b/lib/dry/validation/type_specs.rb
@@ -10,6 +10,34 @@ module Dry
         end
       end
 
+      def build_inherited_type_map(type_specs, inherited_specs)
+        build_type_map(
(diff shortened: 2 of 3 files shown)


alexaltenhuber:
any updates on this PR? any requested changes before this could be merged?

artofhuman:
Many hours spended to digging similar problem :). It would be nice if this PR merged. 
@flash-gordon Do you have a time for review?

thefliik:
See issue #406. I don't think they're accepting PR / maintaining this version of dry-validation.

solnic:
This was addressed in dry-schema so the issue will be gone in dry-validation 1.0.0.

PR #360, opened 2017-08-29
Change missing error message to more useful

olehpratsko:


solnic:
I know it's been ages since this was opened, but now it's outdated as messages api was moved to dry-schema. We can improve this error message there.

PR #370, opened 2017-10-29
Suppress interpolation warnings

flash-gordon:
This is a workaround which suppresses warnings issued by Kernel.sprintf in
presence of extra keys when interpolating a template string. Now, dry-v won't
pollute other projects with a ton of useless warnings. I did some research
on the subject and found out that:

1. The current behavior of MRI is not going to change anytime soon
  https://bugs.ruby-lang.org/issues/2709
  https://bugs.ruby-lang.org/issues/5763

2. What i18n did about this
  a) ignored warnings
  b) implemented interpolation in pure Ruby
  c) added efficient interpolation compiler (this is kinda unrelated but since
     it improves the performance we can have it too to compensate the slowdown
     caused by implementing "sprintf" in Ruby)

  https://github.com/svenfuchs/i18n/issues/65
  https://github.com/svenfuchs/i18n/blob/0a0fa8e5e7f4497e0c39c046b2a94470caf3eb3a/lib/i18n/backend/interpolation_compiler.rb

This commit is step 2a, we then should take 2b and decide whether we need 2c or
not (because we could just use i18n instead).

/cc @solnic @jodosha

### DIFF
--- a/lib/dry/validation/message_compiler.rb
+++ b/lib/dry/validation/message_compiler.rb
@@ -160,6 +160,8 @@ module Dry
       end
 
       def message_text(rule, template, tokens, opts)
+        original_verbosity = $VERBOSE
+        $VERBOSE = nil
         text = template % tokens
 
         if full?
@@ -168,6 +170,8 @@ module Dry
         else
           text
         end
+      ensure
+        $VERBOSE = original_verbosity
       end
 
       def message_tokens(args)
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -30,6 +30,7 @@ end
 
 RSpec.configure do |config|
   config.disable_monkey_patching!
+  config.warnings = true
 
   config.after do
     if defined?(I18n)


solnic:
I'd say fix it however, and we can focus on performant solution for 1.0.0

jodosha:
@flash-gordon @solnic It's sad we have to fix this problem with a hack. Given also it's a problem that other libs ran into. Thanks for taking care of this.

solnic:
oh, I actually didn't notice how it was "fixed" :D yeah this is not nice, we could just reject unknown keys to fix it properly, it'll slow down message processing a bit, but I don't think it's gonna be a big deal for the time being. WDYT?

jodosha:
Even if this fix it's not ideal, as the Ruby VM shouldn't emit these warnings at all, I think this PR it's the most pragmatic solution: no warnings, no slow downs, no extra code to maintain.

flash-gordon:
In principle, compiling templates that interpolate values should be faster than VM's interval code so it's a good feature to have:

```ruby
require "benchmark/ips"

FAST_TEMPLATE = -> (v) {
  "1: #{ v[:first] }, 2: #{ v[:second] }, 3: #{ v[:third] }"
}

SLOW_TEMPLATE = "1: %{first}, 2: %{second}, 3: %{third}"

VALUES = { first: "one", second: "two", third: "three", fourth: "four" }

def fast
  FAST_TEMPLATE.(VALUES)
end

def slow
  SLOW_TEMPLATE % VALUES
end

Benchmark.ips do |x|
  x.report("fast") { fast }
  x.report("slow") { slow }
  x.compare!
end
```

```
                fast    98.721k i/100ms
                slow    51.322k i/100ms
                fast      1.358M (± 6.1%) i/s -      6.812M in   5.038067s
                slow    624.846k (± 4.5%) i/s -      3.131M in   5.020425s

Comparison:
                fast:  1357504.1 i/s
                slow:   624845.5 i/s - 2.17x  slower
```

jodosha:
@flash-gordon Interesting. Are you suggesting to use this approach?

PR #387, opened 2018-02-07
Coercible types can be used to validate

faucct:
Hello. I was searching for a way to have a coercing sanitizing schema. Added specs have the configuration I expected to work.

### DIFF
--- a/lib/dry/validation/input_processor_compiler.rb
+++ b/lib/dry/validation/input_processor_compiler.rb
@@ -24,6 +24,14 @@ module Dry
         send(:"visit_#{node[0]}", node[1], *args)
       end
 
+      def visit_safe(node, *)
+        node.first
+      end
+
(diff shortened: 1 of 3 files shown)


faucct:
Can someone take a look?

thefliik:
See #406, I don't believe they are currently accepting PR for this library.

solnic:
This API was removed in favor of type specs that are now always used in dry-schema, this gem will be used as the schema backend in dry-validation 1.0.0.

solnic:
> See #406, I don't believe they are accepting PR for this library.

We do accept PRs, unfortunately it was decided to rewrite it on top of dry-schema (https://discourse.dry-rb.org/t/plans-for-dry-validation-dry-schema-a-new-gem/215) and due to lack of time maintaining 0.x was not possible.

PR #391, opened 2018-03-07
Fixed bug with nested input for custom validation block

r-glebov:
This is a small fix for nested values in a custom validation block. 

```
require(:newsletter).value(:bool?)
require(:user).schema do
  require(:email).maybe(:str?)
end

validate(email_required: [:newsletter, [:user, :email]]) do |newsletter, email|
  ...
end
```

The `validate` method didn't process the path(as array of symbols) correctly for the nested values.


solnic:
This API was removed in master, its improved version will be available in 1.0.0.

PR #392, opened 2018-03-13
Add indexes to messages when array input checked by hi-level rules

kosmatov:
Related to issue https://github.com/dry-rb/dry-validation/issues/305


solnic:
This API was removed eventually in favor if a simplified `rule` DSL where you can explicitly pass a message string or a locale identifier. It will be available in 1.0.0.

Issue #395, opened 2018-04-13
Incorrect arrays validations in Dry::Struct attributes when used in schemas


semenovDL:
Arrays inside Dry::Struct attributes works incorrectly when used in validation schemas.

```ruby
require 'dry-types'
require 'dry-struct'
require 'dry-validation'

Dry::Validation.load_extensions(:struct)

class S < Dry::Struct
  attribute :amounts, Dry::Types['strict.array'].of(Dry::Types['strict.int'])
end

schema = Dry::Validation.Schema do
  required(:plain).schema do
    required(:amounts).each(Dry::Types['strict.int'])
  end
  required(:struct).schema(S)
end

puts schema.call(
  plain: { amounts: [] },
  struct: { amounts: [] }
).errors

# expected: 
# {}
# actual:
# {:struct=>{:amounts=>["must be filled"]}}

puts schema.call(
  plain: { amounts: ['wrong', 123] },
  struct: { amounts: ['wrong', 123] }
).errors

# expected:
# {:plain=>{:amounts=>{0=>["must be Integer"]}}, :struct=>{:amounts=>{0=>["must be Integer"]}}}

# actual:
# {:plain=>{:amounts=>{0=>["must be Integer"]}}}
```

semenovDL:
Fix it in [#396](https://github.com/dry-rb/dry-validation/pull/396)

solnic:
Support for dry-struct (at least in this form) will be removed in 1.0.0 so this won’t be addressed.

PR #396, opened 2018-04-13
Fix incorrect arrays validations in Dry::Struct attributes when used in schemas

semenovDL:
This PR fix problem with array validation inside Dry::Struct [#395](https://github.com/dry-rb/dry-validation/issues/395)

### DIFF
--- a/lib/dry/validation/extensions/struct.rb
+++ b/lib/dry/validation/extensions/struct.rb
@@ -7,7 +7,13 @@ module Dry
         def create_class(target, other = nil)
           if other.is_a?(Class) && other < Dry::Struct
             super do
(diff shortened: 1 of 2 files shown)


jbielick:
Also running into this issue.

```ruby
class Thing < Dry::Struct::Value
  module Types
    include Dry::Types.module
  end

  input input.strict

  attribute :names, Types::Strict::Array.of(Types::Strict::String)
end

schema = Dry::Validation.Params do
  required(:thing).filled(Thing)
end; nil

schema.(thing: {names: [1]})

# => #<Dry::Validation::Result output={:thing=>{:names=>[1]}} errors={}>
```

Last line should error that `names.0` is not a string.

solnic:
This extension was removed so this is no longer applicable.

PR #400, opened 2018-04-26
Add option to return message_set as failure case in monads extension

jeandenis-k:
Hi! I needed access to the underlying message_set in the failure case of the Result monad, so I made this quick change. Is this adequate?

PS: By the way I noticed the monads extension isn't documented, is this something I could work on?


jeandenis-k:
Given that maintenance is focused on the 1.0 version (according to #406), I'm closing my pull request. Good luck to the team for the 1.0!

Issue #367, opened 2017-10-04
The confirmation macro always reports an error if the schema is defined with type_specs


daniels:
Example without and with type specs:

```ruby
schema1 = Dry::Validation.Form do
  required(:pin).filled(:str?).confirmation
end

schema1.call(pin: "1234").errors
#=> {:pin_confirmation=>["must be equal to 1234"]}

schema1.call(pin: "1234", pin_confirmation: "1234").errors
#=> {}

schema2 = Dry::Validation.Form do
  configure { config.type_specs = true }
  required(:pin, :string).filled.confirmation
end

schema2.call({pin: "1234"}).errors
#=> {:pin_confirmation=>["must be equal to 1234"]}

schema2.call(pin: "1234", pin_confirmation: "1234").errors
#=> {:pin_confirmation=>["must be equal to 1234"]}
```

solnic:
Closing because `confirmation` is going away in the 1.0.0 rewrite

PR #403, opened 2018-05-09
Confirmation with type specs

daniels:
An attempt to solve my own issue with confirmations not working with type_specs. (#367)

Please note that I have little knowledge of what states `type_map[name]` can be in – this is the way I saw to get access to the type_spec of the current rule, there might be better/safer ways! No specs seems to be failing though.


solnic:
This API is gone now. Type specs are now always used in dry-schema that will be the schema engine in dry-validation 1.0.0.

PR #415, opened 2018-08-06
Custom predicates with more than 2 arguments

fran-worley:
@solnic I know that you plan to replace the current implementation of dry-validation but I figured more integration tests and a bug fix might still be worth while!


solnic:
this API is gone now

PR #434, opened 2019-01-10
Builds better schema rules from Dry::Struct (StructClassBuilder)

jbielick:
Struct Extensions (`Dry::Validation.load_extensions(:struct)`)

Dry::Struct can be used in schema validation with the `.schema(`, `.filled(` method to define a nested structure that should validate (satisfy) a predefined Dry::Struct.

The issue addressed here is that there are some assumptions baked into the StructClassBuilder that has several unintuitive behaviors:

https://github.com/dry-rb/dry-validation/blob/88cb4fb38add9abecfee4a5bae09d2a07fd90e80/lib/dry/validation/extensions/struct.rb#L6-L16

1. all keys are declared as `required`, which ignores the meta[:omittable] metadata declared on the struct field type
2. all keys' values are assumed as `.filled`, which may not be the case and ignores the Struct field's Type coercible and `optional` settings
3. Structs with Array-type fields don't validate the array members properly, as noted in #396 (https://github.com/dry-rb/dry-validation/pull/396)
4. Structs with Array-type fields aren't allowed to be empty, which a Struct allows.

I'm in agreement about not using structs to validate, but it seems like this functionality was at least somewhat intended as the extension exists today. If this is unwanted there's already a workaround--define your schema validation separately from your struct. Doing so means
a good deal of duplicating the type information between the schema and the struct.

This modifies the StructClassBuilder to build a schema more intelligently, by respecting `omittable` and handling arrays of types and other nested structs. The time complexity seems similar in that this
is as recursive as the previous implementation, with the exception that this has a few easy conditionals.

### DIFF
--- a/lib/dry/validation/extensions/struct.rb
+++ b/lib/dry/validation/extensions/struct.rb
@@ -5,13 +5,36 @@ module Dry
     class Schema
       module StructClassBuilder
         def create_class(target, other = nil)
-          if other.is_a?(Class) && other < Dry::Struct
+          _self = self
+          if struct?(other)
             super do
-              other.schema.each { |attr, type| required(attr).filled(type) }
+              other.schema.each do |attr, type|
+                chain = if type.meta[:omittable]
+                  optional(attr)
+                else
+                  required(attr)
+                end
+
+                if _self.struct?(type)
+                  chain.schema(type)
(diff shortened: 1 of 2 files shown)


jbielick:
Closing because there seems to be no activity here.

PR #440, opened 2019-02-04
Rewrite

solnic:
This is the 1.0.0 rewrite on top of dry-schema and other dry-* friends. This PR implements core functionality, further improvements, and additional features will be done separately in other PRs. The main goal of this PR is to figure out the API and make basic features work.

### New API (WIP)

``` ruby
class NewUserContract < Dry::Validation::Contract
  params do
    required(:email).filled(:string)
    required(:age).filled(:integer)
    optional(:login).maybe(:string, :filled?)
    optional(:password).maybe(:string, min_size?: 10)
    optional(:password_confirmation).maybe(:string)
  end

  rule(:password) do
    if params[:login] && !params[:password]
      failure("is required")
    end
  end

  rule(:age) do
    if params[:age] < 18
      failure("must be greater or equal 18")
    end
  end
end
```

### TODO

- [x] Figure out the final API (feel free to chime in and discuss here)
- [x] Make travis pass
- [x] Introduce new API for high-level rules: 
  - [x] Make setting `failure` messages work
  - [x] Add support for DI
  - [x] Expose contract's methods within "rule" blocks
  - [x] Expose params in rule blocks
  - [x] Add support for anonymous rules
  - [x] ~~Figure out how to handle failure identifiers, ie `failure(:name, "message")` or set `:name` in the rule method and just use `failure("message")`?~~ so, I ended up doing `rule(identifier)` because I'll move reading values from params to the result object itself on dry-schema side.
- [x] Make `Contract#call` return an actual result object compatible with dry-schema's results
- [x] Make inheritance work with contract classes
- [x] Add support for messages system (YAML, I18n)

### Follow-up TODO (after this is merged)

- [ ] (dry-schema) Extend `Result#error?` to support nested keys ie `error?([:foo, :bar])` or even `error?("foo.bar")` etc. (needs to support arrays too)
- [ ] Improve `Messages` system (use callable templates instead of strings with token placeholders)
- [ ] Add support for dependant rules ie rule B depends on the success of rule A


lib/dry/validation/evaluator.rb
```diff
@@ -0,0 +1,53 @@
+require 'dry/initializer'
```
flash-gordon: I don't see dry-initializer being a dependency.

solnic: Thanks, fixed

lib/dry/validation/.#contract.rb
```diff
@@ -0,0 +1 @@
+solnic@helvete-mbp.local.10586
```
flash-gordon: this is a something's leftover

lib/dry/validation/contract/class_interface.rb
```diff
@@ -0,0 +1,34 @@
...
+        end
+
+        def rule(name, &block)
+          rules << Rule.new(name: name, block: block)
+          rules
+        end
+
+        def schema
+          if defined?(@__schema__)
+            instance_variable_get('@__schema__')
```
flash-gordon: it's just `@__schema__`, `instance_variable_get` is redundant

lib/dry/validation/contract/class_interface.rb
```diff
@@ -0,0 +1,34 @@
...
+          end
+        end
+
+        def rules
+          if defined?(@__rules__)
+            @__rules__
+          else
+            @__rules__ = EMPTY_ARRAY.
```
flash-gordon: why not `||=`? I'm not being nitpicky, just sincerely curious

lib/dry/validation/result.rb
```diff
@@ -1,83 +1,39 @@
...
-      def messages(options = EMPTY_HASH)
-        message_set(options).dump
-      end
+      attr_reader :errors
 
-      def errors(options = EMPTY_HASH)
-        message_set(options.merge(hints: false)).dump
+      def initialize(params, errors)
+        @params = params
+        @errors = errors.update(params.errors)
```
flash-gordon: this mutates input value, is it safe?

lib/dry/validation/result.rb
```diff
@@ -1,83 +1,39 @@
...
 
-      def result_ast
-        @result_ast ||= results.map(&:to_ast)
+      def update(new_errors)
+        errors.update(new_errors)
       end
     end
   end
-end
+end
```
flash-gordon: something's changed in your editor settings :)

lib/dry/validation/result.rb
solnic: @flash-gordon yeah it's being built via `.new` and then it's frozen. See how it's used in `Contract#call`

solnic:
@flash-gordon thank you <3 I addressed your feedback!

solnic:
This was merged BUT the final API is still a WIP. There are lots of things to address and doing it in separate PRs makes much more sense, that's why it was important to get this gigantic PR into master already.

PR #450, opened 2019-02-27
Implemented .build shortcut

igor-alexandrov:
Added `.build` to `Dry::Validation::Contract`.
It is just a shortcut that can be used to define contracts that won't be reused or inherited.


solnic:
Could you rebase this so we can see a green build? 😄

igor-alexandrov:
This passed too: https://travis-ci.org/dry-rb/dry-validation/builds/500297323

Issue #444, opened 2019-02-26
Add YARD docs


solnic:

igor-alexandrov:
@solnic I guess, this can be closed?

PR #452, opened 2019-02-28
Issue #444

igor-alexandrov:
Added basic YARD documentation and update Rule interface to be able to define rule names with String.


solnic:
I added you as a collaborator, so please create branches here :) I also cherry-picked yard-related commit and pushed it to master, rebased this on top of master and opened #453 to replace this PR.

igor-alexandrov:
OK

PR #453, opened 2019-03-01
Updated Rule interface to accept names as String

solnic:


solnic:
So, this is expected to work, but we don't want to coerce to symbols. There are two things to consider here. One is the rule name (unique identifier), which can be used to retrieve error message, and the other thing is data path identifier that we use to check if values are valid before applying rules. Right now rule identifier == data path identifier, but eventually we want to separate the two, even if only one value is passed to `rule` method.

I imagine something like this to work too (eventually):

``` ruby
rule(:unique_email, 'address.email') do
  failure unless unique?(values[:address][:email])
end
```

This sets the rule identifier to `:unique_email` and sets `address.email` as the path to the value that needs to be checked before applying this rule.

igor-alexandrov:
So closing this and starting new issue to support functionality, described above?

solnic:
Yeah let's close this but hold off a bit with further work. I'd like to talk this through with Tim too. I'll report a new issue describing how this API should look like. We could further discuss it there as well. OK?

PR #456, opened 2019-03-02
Raise error if schema definition is duplicated

igor-alexandrov:


lib/dry/validation/contract/class_interface.rb
```diff
@@ -17,7 +17,8 @@ module ClassInterface
         #
         # @api public
         def params(&block)
-          @__schema__ ||= Schema.Params(schema_opts, &block)
+          raise ::Dry::Validation::DuplicateSchemaError, 'Schema has alredy been defined' if defined?(@__schema__)
+          @__schema__ = Schema.Params(schema_opts, &block)
```
solnic: it'd be good to avoid the duplication and introduce a common method, something like `define(:Params, &block)` and handle this check / raising there.

igor-alexandrov:
@solnic  I moved it to separate method. I used `switch`, since I don't want to have any meta-programming in private apis.

lib/dry/validation/contract/class_interface.rb
```diff
@@ -82,15 +86,26 @@ def messages
...
+        private
 
         # @api private
         def schema_opts
           { parent: superclass&.__schema__, config: config }
         end
+
+        # @api private
+        def define(method_name, &block)
+          raise ::Dry::Validation::DuplicateSchemaError, 'Schema has alredy been defined' if defined?(@__schema__)
```
solnic: `alredy` => `already`

PR #488, opened 2019-03-18
Remove Result#to_hash

flash-gordon:
Having it defined in classes that are not completely Hash-like is a common source of problems due to implicit conversions made by ruby when passing keywords around. Current ruby behavior may change in the future but for now it seems it's better not to define `to_hash` without clear benefits.


timriley:
Strongly in support of this 👍

PR #490, opened 2019-03-19
Fix generating messages from symbols for symbol paths

flash-gordon:


lib/dry/validation/evaluator.rb
```diff
@@ -92,7 +92,8 @@ def initialize(*args, &block)
       #
       # @api public
       def key(path = self.path)
-        (@key ||= EMPTY_HASH.dup)[path] ||= Failures.new(path)
+        key = Array(path)
```
solnic: @flash-gordon this ain't right, unfortunately, because it could be a nested hash, too, in which case it'd break. Why did that fix the issue you found?

flash-gordon: @solnic I'm actually figuring out if it did. At least a test is there, does it look right to you?

solnic: @flash-gordon the test looks right. messages backend expects `path` to a message template to be an array, that's why it crashes from what I understand. I believe the right place to coerce it to an array will be `Messages::Resolver#call`.

flash-gordon: @solnic ok, later today I'll try to write a failing test for the case you mentioned. Fixing shouldn't be a problem

flash-gordon: @solnic now I'm recalling what I had in mind, for some reason I decided `key([:email])` is allowed and in this case it should be the same as `key(:email)`, but now looking at the code I don't see any evidence for my belief :)

flash-gordon: @solnic it's late so forgive me in advance
I have a schema
```ruby
class Schema < Dry::Validation::Contract
  schema do
    required(:foo).filled
  end

  rule(:foo) do
    key(bar: :baz).failure('always')
  end
end
```
applying it with dry-v _before_ my changes gives the following results
```
Schema.new.({foo: 1})
 => #<Dry::Validation::Result{:foo=>1} errors={[:bar, :baz]=>["always"]}>
2.6.1 :005 > Schema.new.({foo: 1}).errors.to_h
 => {[:bar, :baz]=>["always"]} 
```
is this expected?

Seems like we should have more cases for explicit paths covered...

solnic: no, this is not how it should work, hashes should produce nested errors

PR #493, opened 2019-03-22
Revert "Sync with schema changes in dry-types"

flash-gordon:
Reverts dry-rb/dry-validation#443


rafaelfranca:
Any reason for this? `dry-types` 0.14 doesn't work in my app due to https://github.com/dry-rb/dry-types/pull/287 and that is only fixed in 0.15.

flash-gordon:
Yes, there are quite specific reasons, dry-types doesn't provide the API required for dry-validation < 1.0 anymore so it's not possible to use these two gems together. I cut a branch for dry-types 0.14 and cherry-picked the fix you referenced. I'll push 0.14.1 shortly.

flash-gordon:
https://rubygems.org/gems/dry-types/versions/0.14.1

rafaelfranca:
Thank you so much ❤️

PR #501, opened 2019-04-01
CI: use 2.4.6

olleolleolle:
This PR updates the CI matrix.


waiting-for-dev:
If I'm not wrong, setting just `2.4` will make travis fetch last point release version. We could do that for all ruby versions in all `dry` repos, so we don't have to manually change it at every release. What do you think?

solnic:
@waiting-for-dev yes if you can confirm that this is how it works. I tried that some time ago and **it did not work like that**

waiting-for-dev:
I recently did it in a project ([.travis.yml](https://github.com/waiting-for-dev/warden-jwt_auth/blob/master/.travis.yml), [build](https://travis-ci.org/waiting-for-dev/warden-jwt_auth)) and it worked, but I haven't seen it documented.

solnic:
We need to confirm that with travis folks.

olleolleolle:
The file in `rvm` which lists what aliases are active: https://github.com/rvm/rvm/blob/master/config/known

If `rvm` is updated, and using that file, things will work right.

waiting-for-dev:
Good research! :)

Isn't rvm using that file **always** ? If so, then we would only rely on travis keeping rvm updated, which I don't know if they do very often. Anyway, the worst case scenario would be just testing against older versions in the same minor version branch, which I think it's not too terrible and it would unload a quite boring maintenance burden.

PR #515, opened 2019-04-25
Fix: Contract example class name and missing `key.failure`

gcentauri:


gcentauri:
@solnic - Would it be worth pulling this example directly into the dry-rb docs repo?  I'm slowly finding time to work on this little by little.

solnic:
@gcentauri yes! We need a set of new user docs on dry-rb.org

PR #525, opened 2019-05-14
Display the whole path for errors

Morozzzko:
Before:

I had a code that looked like this:

```ruby
rule do
  base.failure(:expired)
end
```

which returned this error:

```
       Message template for :expired under "" was not found
```

I figured it would be nice to be a bit more specific and probably display *where* the message was supposed to be. 

Now it says

```
       Message template for :expired under ["dry_validation.rules.expired"] was not found
```

Also, found out that the message's `path` param does not really affect the i18n rule path.

**Beware** I've used Dry::Schema::Message#rule_lookup_paths` which is marked by `@api private`. Probably should be extra considerate about that.

### DIFF
--- a/lib/dry/validation/messages/resolver.rb
+++ b/lib/dry/validation/messages/resolver.rb
@@ -68,7 +68,7 @@ module Dry
 
           unless template
             raise MissingMessageError, <<~STR
-              Message template for #{rule.inspect} under #{keys.join(DOT).inspect} was not found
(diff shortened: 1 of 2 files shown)


Morozzzko:
Actually, it hasn't fixed my problem, but still

Morozzzko:
Seems like I'll need to come up with another fix, as this one does not seem to be adequate.

But I'd like to see a meaningful error message

PR #526, opened 2019-05-14
[WIP] [FIX] I18n fails for symbolized base rule failures

Morozzzko:
```
rule do
  base.failure(:foo)
end
```

Fails with `       Message template for :not_good_enough under "" was not found` even if the locale string is present.

I'm fixing it at the moment. Any suggestions are welcome

### DIFF
--- a/spec/fixtures/messages/errors.en.yml
+++ b/spec/fixtures/messages/errors.en.yml
@@ -2,6 +2,7 @@ en:
   dry_validation:
     rules:
       email: "E-mail"
+      not_good_enough: "Email should be nicer"
     errors:
       rules:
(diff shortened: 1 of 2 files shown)


Morozzzko:
Apparently, it will work if we use `dry_validation.errors.rules.not_good_enough` instead of `dry_validation.rules.not_good_enough`.

I'm confused and I need some help. Should we go for a fix and allow us to place the thing under `dry_validation.rules` instead of `dry_validation.errors.rules`?

Morozzzko:
Welp, I figured I was wrong the whole time, and everything around was designed to use `errors.rules` 🤦‍♂ 

Sorry.

Issue #538, opened 2019-06-10
each + macro without args broken


mensfeld:
won't work:

```ruby
rule(:seed_brokers).each(:broker_schema)
```

works:

```ruby
rule(:seed_brokers).each(broker_schema: true)
```

jandudulski:
it works for me :thinking:

solnic:
```ruby
# frozen_string_literal: true

require 'dry/validation'

class MyContract < Dry::Validation::Contract
  params do
    required(:num).filled(:integer)
  end

  register_macro(:even?) do
    key.failure('must be even') unless value.even?
  end

  rule(:num).validate(:even?)
end

contract = MyContract.new

puts contract.call(num: 3).errors.inspect
# #<Dry::Validation::MessageSet messages=[#<Dry::Validation::Message text="must be even" path=[:num] meta={}>] options={}>

puts contract.call(num: 4).errors.inspect
# #<Dry::Validation::MessageSet messages=[] options={}>
```

I can't reproduce this either.

mensfeld:
```ruby
# frozen_string_literal: true

require 'dry/validation'

class MyContract < Dry::Validation::Contract
  params do
    required(:nums).filled(:array)
  end

  register_macro(:even?) do
    key.failure('must be even') unless value.even?
  end

  rule(:nums).each(:broker_schema?)
end

contract = MyContract.new

puts contract.call(nums: [3]).errors.inspect

 undefined method `flatten' for :broker_schema?:Symbol (NoMethodError)
```

solnic:
oh doooh this one is about `each` sorry. got it reproduced, ty, will fix.

mensfeld:
same for:

```ruby
# frozen_string_literal: true

require 'dry/validation'

class MyContract < Dry::Validation::Contract
  params do
    required(:nums).value(:array, :filled?)
  end

  register_macro(:even?) do
    key.failure('must be even') unless value.even?
  end

  rule(:nums).each(:even?)
end

contract = MyContract.new

puts contract.call(nums: [3]).errors.inspect
```

Issue #541, opened 2019-06-11
Rule `each` doesn't work with list of macros


jandudulski:
```ruby
RSpec.describe "Contract" do
  subject(:contract) do
    Dry::Validation.Contract do
      register_macro(:foo) do
        key.failure("not a foo") unless value == "foo"
      end

      register_macro(:bar) do
        key.failure("not a bar") unless value == "bar"
      end

      params do
        required(:test).array(:string)
      end

      rule(:test).each(:foo, :bar)
    end
  end

  it "validates each" do
    res = contract.call(test: ["foo", "bar"])

    expect(res.errors.to_h).to eql(test: {0 => ["not a bar"], 1 => ["not a foo"]})
  end
end
```

This fails with error:

```
     NoMethodError:
       undefined method `flatten' for :foo:Symbol
```

Also when I change to `rule(:test).each([:foo, :bar])` the result is not as expected as well:

```
       Diff:
       @@ -1,2 +1,2 @@
       -:test => {0=>["not a bar"], 1=>["not a foo"]},
       +:test => {1=>["not a foo"]},
```

jandudulski:
I would expect this to work since `rule(:foo).validate(:one, :two)` works as expected.

jandudulski:
Oh, actually `rule(:foo).validate(:one, :two)` this doesn't work as well - it just evaluates the first macro - tests added #543

PR #543, opened 2019-06-11
Fix for handling multiple macros

jandudulski:


solnic:
@jandudulski 👋 are you planning to fix it?

jandudulski:
@solnic I'm trying but tbh there are a lot of moving parts here and some of them are mysterious to me. Like this:

```ruby
# evaluator
macros.each do |args|
  macro = macro(*args.flatten(1))
  instance_exec(macro.extract_block_options(_options.merge(macro: macro)), &macro.block)
end
```

or this

```ruby
# rule
@macros = macros.map { |spec| Array(spec) }.map(&:flatten)
```

so it might take some time before I will be able to solve this. Any tips here are more than welcome :D

solnic:
@jandudulski you found the right places where things are broken. We need to make sure that `@macros` in `Rule` is set in the following way:

1) `validate(:foo)` => `[:foo]`
1) `validate(:foo, :bar)` => `[:foo, :bar]`
1) `validate(foo: 1)` => `[[:foo, [1]]]`
1) `validate(foo: [1,2])` => `[[:foo, [1,2]]]`
1) `validate(foo: [1], bar: [2])` => `[[:foo, [1,2]], [:bar, [2]]]`

and then that these cases are handled properly in `Evaluator#initialize` where we pass those to `Contract#macro`

jandudulski:
@solnic that's something I can work with.

jandudulski:
@solnic please take a look

jandudulski:
Fixes #538 and #541

spec/unit/rule_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      expect(rule.parse_macros(foo: %i[bar baz]))
+        .to eql([[:foo, %i[bar baz]]])
+    end
+
+    it 'works with multiple hashes' do
+      expect(rule.parse_macros(foo: [:bar], baz: [:qux]))
+        .to eql([[:foo, [:bar]], [:baz, [:qux]]])
+    end
+  end
+end
```
solnic: @jandudulski could you actually remove this spec and test these cases in an integration spec for `validate` rule method? this will give more confidence, plus `parse_macros` feels more like an implementation detail rather than a public API that should be directly tested.

jandudulski: I'll do it later today I hope

Issue #537, opened 2019-06-10
Import dry-logic predicates as macros


solnic:
The idea: import most dry-logic predicates as macros

## PoC

```ruby
require 'dry/validation'

class ApplicationContract < Dry::Validation::Contract
end

predicates = Dry::Logic::Predicates.public_methods.grep(/\?\Z/).sort - Object.public_methods - Module.public_methods
registry = Dry::Schema::PredicateRegistry.new(Dry::Logic::Predicates)

predicates.each do |name|
  arg_list = registry.arg_list(name).map(&:first)

  ApplicationContract.register_macro(name) do |macro:|
    args = [*macro.args, value]
    result = Dry::Logic::Predicates.public_send(name, *args)

    return if result

    message_opts = arg_list.zip(args).to_h

    key.failure(name, message_opts)
  end
end

class MyContract < ApplicationContract
  params do
    required(:age).filled(:integer)
  end

  rule(:age).validate(gteq?: 18)
end

contract = MyContract.new

puts contract.(age: 17).errors.to_h.inspect

# {:age=>["must be greater than or equal to 18"]}
```

## Requirements

* This should exclude *some* predicates, like type-checks and maybe some more
* This should be an extension, so that users can load it on demand via `Dry::Validation.load_extensions(:predicate_macros)`
* The extension should add a contract class method that will load macros, we don't want to define them as global macros after all

waiting-for-dev:
@solnic could I take this one?

solnic:
@waiting-for-dev sure, thank you. I just sent you an invite to the repo. Please assign to yourself when it's accepted so that others can see you're working on it.

PR #548, opened 2019-06-13
Extension to use dry-logic predicates as macros

waiting-for-dev:
PoC for `:gteq?`

This is the implementation for #537 

Some considerations:

- Right now it just loads with `:gteq?` predicate. We have to decide which predicates need to be imported. My proposal is:

  - empty?
  - filled?
  - odd?
  - even?
  - lt?
  - gt?
  - lteq?
  - gteq?
  - size?
  - min_size?
  - max_size?
  - inclusion?
  - exclusion?
  - included_in?
  - excluded_from?
  - includes?
  - excludes?
  - eql?
  - is?
  - not_eql?
  - true?
  - false?
  - uuid_v4?
  - respond_to?

- I think `Dry::Validation::Predicate#arg_names` and `Dry::Validation::Predicate#call` should be moved to `Dry::Schema::PredicateRegistry`.

- I think it would be nice to keep `name` and `options` in the generated error when doing `.failure(name, options)`. If I'm understanding what it is happening, they are just used as an intermediate info to build the error message (`must be greater than...` in this case). This way I could test what I want to test (that I provide that information) and not have to rely on the message generation, which makes me depend on the yaml, etc..

- Surely it has been discussed before, but I feel it is a duplication being able to use predicates in `dry-schema`. I think that `dry-schema` should only be concerned with structure and types, while `dry-validation` should deal with values.

Thanks for any feedback.

### DIFF
--- a/lib/dry/validation/extensions/predicates_as_macros.rb
+++ b/lib/dry/validation/extensions/predicates_as_macros.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
+require 'dry/logic/predicates'
+require 'dry/schema/predicate_registry'
+require 'dry/validation/contract'
+
 module Dry
   module Validation
     # Encapsulation of a dry-logic predicate.
--- a/lib/dry/validation/extensions/predicates_as_macros.rb
+++ b/lib/dry/validation/extensions/predicates_as_macros.rb
@@ -11,7 +11,10 @@ module Dry
       # List of predicates to be imported.
       #
       # @see Dry::Validation::Contract
-      WHITELIST = %i[gteq?].freeze
+      WHITELIST = %i[
+        filled? gt? gteq? included_in? includes? inclusion? is? lt?
+        lteq? max_size? min_size? not_eql? odd? respond_to? size? true?
+      ].freeze
 
       # @api private
       REGISTRY = Dry::Schema::PredicateRegistry.new(
--- a/lib/dry/validation/extensions/predicates_as_macros.rb
+++ b/lib/dry/validation/extensions/predicates_as_macros.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require 'dry/logic/predicates'
 require 'dry/schema/predicate_registry'
 require 'dry/validation/contract'
 
--- a/lib/dry/validation/extensions/predicates_as_macros.rb
+++ b/lib/dry/validation/extensions/predicates_as_macros.rb
@@ -14,6 +14,7 @@ module Dry
       WHITELIST = %i[
         filled? gt? gteq? included_in? includes? inclusion? is? lt?
         lteq? max_size? min_size? not_eql? odd? respond_to? size? true?
+        uuid_v4?
       ].freeze
 
       # @api private
--- a/lib/dry/validation/extensions/predicates_as_macros.rb
+++ b/lib/dry/validation/extensions/predicates_as_macros.rb
@@ -6,9 +6,10 @@ require 'dry/validation/contract'
 
 module Dry
   module Validation
-    # Encapsulation of a dry-logic predicate.
-    class Predicate
-      # List of predicates to be imported.
+    # Predicate registry with additional needed methods.
+    class PredicateRegistry < Schema::PredicateRegistry
+      # List of predicates to be imported by `:predicates_as_macros`
+      # extension.
       #
       # @see Dry::Validation::Contract
       WHITELIST = %i[
@@ -18,37 +19,25 @@ module Dry
       ].freeze
 
       # @api private
-      REGISTRY = Dry::Schema::PredicateRegistry.new(
-        Dry::Logic::Predicates
-      ).freeze
-
-      # @api private
-      attr_reader :name
-
-      # @api private
-      def initialize(name)
-        @name = name
+      def arg_names(name)
+        arg_list(name).map(&:first)
       end
 
       # @api private
-      def arg_names
-        REGISTRY.arg_list(name).map(&:first)
+      def call(name, args)
+        self[name].(*args)
       end
 
       # @api private
-      def call(args)
-        REGISTRY[name].(*args)
-      end
-
-      # @api private
-      def message_opts(arg_values)
-        arg_names.zip(arg_values).to_h
+      def message_opts(name, arg_values)
+        arg_names(name).zip(arg_values).to_h
       end
     end
 
     # Extension to use dry-logic predicates as macros.
     #
-    # @see Dry::Validation::Predicate::WHITELIST Available predicates
+    # @see Dry::Validation::PredicateRegistry::WHITELIST Available
+    # predicates
     #
     # @example
     #   Dry::Validation.load_extensions(:predicates_as_macros)
@@ -70,12 +59,12 @@ module Dry
     class Contract
       # Make macros available for self and its descendants.
       def self.import_predicates_as_macros
-        Predicate::WHITELIST.each do |name|
-          predicate = Predicate.new(name)
+        registry = PredicateRegistry.new
+        PredicateRegistry::WHITELIST.each do |name|
           register_macro(name) do |macro:|
             predicate_args = [*macro.args, value]
-            predicate.(predicate_args) ||
-              key.failure(name, predicate.message_opts(predicate_args))
+            registry.(name, predicate_args) ||
+              key.failure(name, registry.message_opts(name, predicate_args))
           end
         end
       end
--- a/lib/dry/validation/extensions/predicates_as_macros.rb
+++ b/lib/dry/validation/extensions/predicates_as_macros.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+module Dry
+  module Validation
+    # Encapsulation of a dry-logic predicate.
+    class Predicate
+      # List of predicates to be imported.
+      #
+      # @see Dry::Validation::Contract
+      WHITELIST = %i[gteq?].freeze
+
+      # @api private
+      REGISTRY = Dry::Schema::PredicateRegistry.new(
+        Dry::Logic::Predicates
+      ).freeze
+
+      # @api private
+      attr_reader :name
+
+      # @api private
+      def initialize(name)
+        @name = name
+      end
+
+      # @api private
+      def arg_names
+        REGISTRY.arg_list(name).map(&:first)
+      end
+
+      # @api private
+      def call(args)
+        REGISTRY[name].(*args)
+      end
+
+      # @api private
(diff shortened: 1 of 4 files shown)


solnic:
> Surely it has been discussed before, but I feel it is a duplication being able to use predicates in dry-schema. I think that dry-schema should only be concerned with structure and types, while dry-validation should deal with values.

That's the recommended way of using dry-validation starting from 1.0.0. We don't want to enforce it though. It's just a tool, do whatever you feel is good for you, including shooting yourself in the foot and learning the lesson 😄

waiting-for-dev:
> That's the recommended way of using dry-validation starting from 1.0.0. We don't want to enforce it though. It's just a tool, do whatever you feel is good for you, including shooting yourself in the foot and learning the lesson 

I see :) However, being it the recommended way of using dry-validation, wouldn't it make sense to have it as core behavior instead of an extension?

solnic:
@waiting-for-dev yes but I want to start with the extension and mark it as experimental. I expect things to fully shape up for 2.0.0. Does this make sense?

waiting-for-dev:
> @waiting-for-dev yes but I want to start with the extension and mark it as experimental. I expect things to fully shape up for 2.0.0. Does this make sense?

Yes, better approach it step by step.

solnic:
> Right now it just loads with :gteq? predicate. We have to decide which predicates need to be imported. My proposal is: (...)

This list looks good to me. Are you planning to continue working on this?

waiting-for-dev:
> This list looks good to me. Are you planning to continue working on this?

Yes, I just wanted some feedback before going forward. I can work on this during this week. What do you think about the other points?:

> I think Dry::Validation::Predicate#arg_names and Dry::Validation::Predicate#call should be moved to Dry::Schema::PredicateRegistry.

I could prepare a PR to `dry-schema` with those two methods added and then continue here.

> I think it would be nice to keep name and options in the generated error when doing .failure(name, options). If I'm understanding what it is happening, they are just used as an intermediate info to build the error message (must be greater than... in this case). This way I could test what I want to test (that I provide that information) and not have to rely on the message generation, which makes me depend on the yaml, etc..

This one is not necessary for this PR, but it could be implemented afterwards.

solnic:
>> I think Dry::Validation::Predicate#arg_names and Dry::Validation::Predicate#call should be moved to Dry::Schema::PredicateRegistry.
> I could prepare a PR to dry-schema with those two methods added and then continue here.

OK, but why is this refactoring needed?

waiting-for-dev:
> > > I think Dry::Validation::Predicate#arg_names and Dry::Validation::Predicate#call should be moved to Dry::Schema::PredicateRegistry.
> > > I could prepare a PR to dry-schema with those two methods added and then continue here.
> 
> OK, but why is this refactoring needed?

It is just that I think that it is where these methods belong. It is a responsibility for `PredicateRegistry`, I'm only dispatching there from `Dry::Schema::PredicateRegistry`. But if you are in a hurry with this issue I can do it afterwards.

waiting-for-dev:
@solnic for now I added missing predicates, so it is complete in that regard.

It seems that `:uuid_v4?` predicate is not in the `dry-logic`'s version in use, so for now I removed it.

I rebased commits to master.

flash-gordon:
Why do we use whitelist in the first place?

solnic:
@flash-gordon can we generate this list in a reliable way?

flash-gordon:
@solnic what predicates should be excluded?

solnic:
@flash-gordon actually, stop, let's keep it hardcoded, we want to document it anyway, having an explicit list will be useful and it's not like we will have to change it all the time. WDYT?

flash-gordon:
but why not give a link to dry-logic where it should be covered anyway?

solnic:
@flash-gordon jumping between docs is a bad experience but I don't want to get into details here (I have an idea how to avoid this)

flash-gordon:
My only concern was about uuid_v4 which is deliberately excluded, we should bump dry-logic instead. It wouldn't be necessary if we didn't have the whitelist.

solnic:
@flash-gordon we can add it now and rely on dry-logic's master in the meantime, then just release new dry-validation along with dry-logic

flash-gordon:
@solnic I released it a week ago or so

solnic:
@waiting-for-dev could you re-add `uuid_v4`? it *is available* in the latest dry-logic (we need to bump min version requirement too)

waiting-for-dev:
> @waiting-for-dev could you re-add uuid_v4? it is available in the latest dry-logic (we need to bump min version requirement too)

@solnic done. I bumped dry-schema minimal version to `1.2`. It is also where dry-logic dependency was bumped for `respond_to?` predicate.

> Why do we use whitelist in the first place?

@flash-gordon my opinion on this is that is better to have a whitelist (explicitness) vs. a blacklist (implicitness). It is possible that another predicate is added to dry-logic and that we don't want it here. The difference is having a feature request from time to time (adding a predicate) vs. having a bug report (a predicate that makes no sense as a validation rule).

flash-gordon:
I'd prefer a bug report and a fix w/o either blacklists or whitelists i.e. by differentiating predicates with a ... predicate.

waiting-for-dev:
> differentiating predicates with a ... predicate.

That would be great. But, how?

flash-gordon:
Determined by the criteria of what predicates we want to keep. First of all, why not have them all, what are the predicates we don't want to have in dry-v? Anyhow, I'm not against current implementation, no need to change anything about it.

waiting-for-dev:
> what are the predicates we don't want to have in dry-v?

We don't want to have as predicates for business validation rules those which are not concerned with data but with the type or structure of that data (like `:string?`). Those predicates have to be used only in the schema definition.

> Determined by the criteria of what predicates we want to keep.

The only way I can think is to differentiate between data predicates and type predicates in dry-logic itself. But from dry-logic's point of view I don't know if it is a good idea.

solnic:
> First of all, why not have them all, what are the predicates we don't want to have in dry-v?

Like @waiting-for-dev said - type checks, because the entire point of having a schema is to NOT have to check for types in the rules. I plan a feature where you'll be able to do pattern-matching in rules that may involve type checks, but that's a whole different story.

lib/dry/validation/extensions/predicates_as_macros.rb
```diff
@@ -0,0 +1,86 @@
...
+      def call(args)
+        REGISTRY[name].(*args)
+      end
+
+      # @api private
+      def add_failure_from_call(key, arg_values)
+        message_opts = arg_names.zip(arg_values).to_h
+
+        key.failure(name, message_opts)
+      end
```
solnic: it would be better to just return `message_opts` here and rename this method to `message_opts` and then in `import_predicates_as_macros` you could just do `key.failure(name, predicate.message_opts)`. Otherwise, `Predicate` will be coupled to evaluator's key API.

waiting-for-dev: Yeah, much better. Done in the last commit.

waiting-for-dev:
> I'm thinking that maybe we want to have Validation::PredicateRegistry < Schema::PredicateRegistry that would implement whatever is needed and be able to leverage the existing API.

Do you mean replacing `Validation::Predicate` by `Validation::PredicateRegirtry < Schema::PredicateRegistry`, alongside moving/adapting its methods? I think that it makes sense for `#message_opts`, but `#arg_names` & `#call` would be in `Validation::PredicateRegistry` just because of circumstantial reasons, as they naturally belong to `Schema::PredicateRegistry` (and who knows if eventually we'll need them in `dry-schema`).

Anyway, tell me what you prefer and I'll do it in next commit.

waiting-for-dev:
> I think that it makes sense for #message_opts

Well, right now I haven't check, but I guess `dry-validation` shares the API of `dry-schema` in this regard. In this case, I think that `#message_opts` should go to `Schema::PredicateRegistry`, too :)

solnic:
> I think that #message_opts should go to Schema::PredicateRegistry

Let's not touch this for now and just start with `Validation::PredicateRegistry`. Dealing with message options is *extremely tricky* on the schema side so I don't want to make any changes related to this as part of this feature.

waiting-for-dev:
Ok, if I understood well, now it should be correct. Because I understand that, for now, we are going to keep `Validation::PredicateRegistry` "local" to the extension. This is also the reason why I define `WHITELIST` constant here and not in `Validation::Contract`. This way I don't pollute further latter namespace.

lib/dry/validation/extensions/predicates_as_macros.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+require 'dry/logic/predicates'
```
solnic: this require is no longer needed

lib/dry/validation/extensions/predicates_as_macros.rb
```diff
@@ -0,0 +1,73 @@
...
+    #   # => 'must be greater than or equal to 18'
+    class Contract
+      # Make macros available for self and its descendants.
+      def self.import_predicates_as_macros
+        registry = PredicateRegistry.new
+        PredicateRegistry::WHITELIST.each do |name|
+          register_macro(name) do |macro:|
+            predicate_args = [*macro.args, value]
+            registry.(name, predicate_args) ||
+              key.failure(name, registry.message_opts(name, predicate_args))
```
solnic: can't we just do:

```ruby
key.failure(name, registry.message_opts(name, predicate_args)) unless registry[name].(*predicate_args)
```

?

waiting-for-dev: I'm sometimes a little fuzzy and prefer using expressions with explicit return values (`if` without `else` returns `nil` on `false` branch, but, wait, is `nil` in any case a value???) :smile: 

Anyway, not a big deal, last commit contains this change. However, rubocop doesn't like it and I should use a guard clause, because a one-line modifier would be too long. Do I change it?

lib/dry/validation/extensions/predicates_as_macros.rb
waiting-for-dev: Good sight :)

lib/dry/validation/extensions/predicates_as_macros.rb
solnic: it's easier to read and understand a conditional (at least for me) and ACTUALLY this should use `or` if we wanted to write more idiomatic Ruby because it's not about returning a value, it's about setting a failure but anyway, a conditional is clearer so let's stick to it.

> Do I change it?

No it's fine, I'll merge and do some quick follow-up work anyway

Issue #540, opened 2019-06-11
Rule fired for optional attribute which is missing


mateusz-useo:
**Description**

Rule is getting fired for a required attribute which is nested within an optional attribute and when the whole optional hash is missing (which should be allowed) it reports an error causing the validation to fail.

**To Reproduce**

Create a contract with a rule for an optional hash attribute with a required key + a rule for the required key and run with an input not including that hash. I use custom macro and I don't know if it's related to the bug.

So given:
```ruby
class SignUpContract < Dry::Validation::Contract
  params do
    required(:user).hash do
      optional(:id) { filled? & str? }
      optional(:first_name) { filled? & str? }
      optional(:last_name) { filled? & str? }
      required(:email) { filled? & str? }
      required(:password) { filled? & str? }
      required(:accepts_terms) { filled? & bool? }
      optional(:freelancer).hash do
        required(:id) { filled? & str? }
      end
      optional(:clients).array(:hash) do
        required(:id) { filled? & str? }
      end
    end
  end

  rule('user.id').validate(:uuid_format)
  rule('user.email').validate(:email_format)
  rule('user.password').validate(min_length: 8)

  rule(user: %i[freelancer clients]) do
    freelancer = values.dig(:user, :freelancer)
    clients = values.dig(:user, :clients)

    unless freelancer ^ (clients && clients.size == 1)
      key(:user).failure(:profile_missing)
    end
  end

  rule('user.freelancer.id').validate(:uuid_format)
  rule('user.clients') do
    have_valid_uuid = lambda { |client|
      client[:id].nil? || client[:id].match?(Macros::UUID_REGEXP)
    }

    unless value.all?(have_valid_uuid)
      key.failure('not a valid UUID format')
    end
  end
end
```
And params:
```json
{
  "user": {
    "id": "d16dfef3-50f9-4588-b7d7-837801e403cd",
    "first_name": "Blah",
    "last_name": "Blah",
    "email": "blah@blah.blah",
    "password": "blahblahblah",
    "accepts_terms": true,
    "clients": [
      {
        "id": "d287fdff-78ac-45ed-871b-b1b5c9548399"
      }
    ]
  }
}
```
It produces:
```ruby
[...]

PR #550, opened 2019-06-14
Add key? method to rule blocks

solnic:
This is helpful when dealing with optional keys. We can't just assume
that we do not want to execute a rule when a value was missing.

Good example: a password must be provided when login is provided, and
both are optional.

Refs #540

### DIFF
--- a/lib/dry/validation/evaluator.rb
+++ b/lib/dry/validation/evaluator.rb
@@ -153,13 +153,13 @@ module Dry
       #
       # @example
       #   rule(:age) do
-      #     key.failure(:invalid) if value? && value < 18
+      #     key.failure(:invalid) if key? && value < 18
       #   end
       #
       # @return [Boolean]
       #
       # @api public
-      def value?
+      def key?
         values.key?(key_name)
       end
 
--- a/lib/dry/validation/evaluator.rb
+++ b/lib/dry/validation/evaluator.rb
@@ -147,6 +147,22 @@ module Dry
         values[key_name]
       end
 
+      # Return if the value under the default key is available
+      #
+      # This is useful when dealing with rules for optional keys
+      #
+      # @example
+      #   rule(:age) do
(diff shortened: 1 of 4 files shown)


lib/dry/validation/values.rb
```diff
@@ -51,6 +51,18 @@ def [](*args)
         end
       end
 
+      # @api public
+      def key?(key)
+        return data.key?(key) if key.is_a?(Symbol)
+
+        Schema::Path[key].reduce(data) do |a, e|
+          return false unless a.key?(e)
```
flash-gordon: this will break if `a` is an array

solnic: @flash-gordon hah I actually just fixed that

lib/dry/validation/evaluator.rb
```diff
@@ -147,6 +147,22 @@ def value
         values[key_name]
       end
 
+      # Return if the value under the default key is available
+      #
+      # This is useful when dealing with rules for optional keys
+      #
+      # @example
+      #   rule(:age) do
+      #     key.failure(:invalid) if value < 18 if value? && value < 18
```
flash-gordon: two `if` seem redundant

lib/dry/validation/evaluator.rb
```diff
@@ -147,6 +147,22 @@ def value
...
+      #   end
+      #
+      # @return [Boolean]
+      #
+      # @api public
+      def value?
```
timriley: I wonder if this is actually the right name? In my mind `key?` actually seems like the more appropriate name? To my instincts, at least, `value?` feels like something that's returning the truthiness of the value...

timriley: If you consider it in usage, I think it feels pretty good named as `#key?` too...

```ruby
      rule(:login) do
         if key?
           key.failure('too short') if value.length < 3
         end
       end
```

PR #568, opened 2019-07-12
Fix each value not being an array

alexandru-calinoiu:
Relates to https://github.com/dry-rb/dry-validation/issues/567

### DIFF
--- a/lib/dry/validation/rule.rb
+++ b/lib/dry/validation/rule.rb
@@ -82,7 +82,7 @@ module Dry
         @keys = []
 
         @block = proc do
-          (values[root] || []).each_with_index do |_, idx|
+          ((values[root] if values[root].is_a?(Array)) || []).each_with_index do |_, idx|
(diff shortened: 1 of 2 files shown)


lib/dry/validation/rule.rb
```diff
@@ -82,7 +82,7 @@ def each(*macros, &block)
         @keys = []
 
         @block = proc do
-          (values[root] || []).each_with_index do |_, idx|
+          ((values[root] if values[root].is_a?(Array)) || []).each_with_index do |_, idx|
```
alexandru-calinoiu: Ideally we should not reach this code if params validation does not pass, but I am not sure how do that yet.

alexandru-calinoiu:
I can deal with the codeclimate issue if you guys let me know if I am on the right path here.

solnic:
Thanks, I ended up fixing it in #572

PR #577, opened 2019-07-31
dry/validations/values/#[] - Use fetch_values on final map

esparta:
Trying to express better the intention, additionally it could be faster on cRuby and jRuby implementations

### DIFF
--- a/lib/dry/validation/values.rb
+++ b/lib/dry/validation/values.rb
@@ -45,15 +45,13 @@ module Dry
 
         case (key = args[0])
         when Symbol, String, Array, Hash
-          path = Schema::Path[key]
-          keys = path.to_a
+          keys = Schema::Path[key].to_a
 
           return data.dig(*keys) unless keys.last.is_a?(Array)
 
           last = keys.pop
           vals = self.class.new(data.dig(*keys))
-
-          last.map { |name| vals[name] }
+          vals.fetch_values(*last)
         else
           raise ArgumentError, '+key+ must be a valid path specification'
         end
--- a/lib/dry/validation/values.rb
+++ b/lib/dry/validation/values.rb
@@ -51,7 +51,7 @@ module Dry
 
           last = keys.pop
           vals = self.class.new(data.dig(*keys))
-          vals.fetch_values(*last)
+          vals.fetch_values(*last) { nil }
         else
           raise ArgumentError, '+key+ must be a valid path specification'
         end
--- a/spec/unit/values_spec.rb
+++ b/spec/unit/values_spec.rb
@@ -42,6 +42,10 @@ RSpec.describe Dry::Validation::Values do
           ArgumentError, '+key+ must be a valid path specification'
         )
     end
+
+    it 'accepts missing keys returning nil' do
+      expect(values[address: { geo: [:population, :lon] }]).to eql([nil, 2])
+    end
   end
 
   describe '#key?' do


lib/dry/validation/values.rb
```diff
@@ -45,15 +45,13 @@ def [](*args)
...
-          keys = path.to_a
+          keys = Schema::Path[key].to_a
 
           return data.dig(*keys) unless keys.last.is_a?(Array)
 
           last = keys.pop
           vals = self.class.new(data.dig(*keys))
-
-          last.map { |name| vals[name] }
+          vals.fetch_values(*last)
```
esparta: Current test doesn't cover what's the expected value if the key doesn't exist, and following the flow looks like it should not be a problem.
If need to have a non-existent key (as nil) this would probably needed.
```suggestion
          vals.fetch_values(*last) { nil }
```

solnic: I think we want to stick to Ruby's (poor, but "idiomatic" convention), that it should return `nil` when a key doesn't exist.

solnic:
Did you compare performance?

esparta:
Yes I did. The code is here: https://gist.github.com/esparta/c905af2a0220daca37cd7adf44db6b0d

```
Case: works with a hash pointing to multiple values
             current    11.088k i/100ms
   original proposal    17.183k i/100ms
   proposal with nil    16.085k i/100ms
             current    118.203k (± 1.2%) i/s -    598.752k in   5.066212s
   original proposal    191.301k (± 1.3%) i/s -    962.248k in   5.030910s
   proposal with nil    182.536k (± 1.4%) i/s -    916.845k in   5.023826s

Comparison:
   original proposal:   191301.4 i/s
   proposal with nil:   182535.6 i/s - 1.05x  slower
             current:   118203.4 i/s - 1.62x  slower

Case: works with a path
             current    39.495k i/100ms
   original proposal    36.411k i/100ms
   proposal with nil    36.908k i/100ms
             current    456.994k (± 1.4%) i/s -      2.291M in   5.013580s
   original proposal    473.574k (± 1.0%) i/s -      2.403M in   5.074950s
   proposal with nil    478.554k (± 1.1%) i/s -      2.399M in   5.013634s

Comparison:
   proposal with nil:   478554.1 i/s
   original proposal:   473574.4 i/s - same-ish: difference falls within error
             current:   456994.1 i/s - 1.05x  slower
```

As expected, the change is more relevant when using complex hashes or even arrays of.

solnic:
Oh, I'm sorry I didn't make myself clear - I meant comparing performance between using `fetch_values` vs the previous implementation.

esparta:
Sorry also about the confusion, perhaps my labels were not clear enough:

- current -> `last.map { |name| vals[name] }`
- original proposal -> `vals.fetch_values(*last)`
- proposal with nil -> `vals.fetch_values(*last) { nil }`

the micro-benchmark suggest that `fetch_values` is 1.6x faster than a ruby map block in the scenario with complex hashes, and same-ish on the other case.

solnic:
OK that’s awesome. Could you also add a spec for the `nil` part please?

esparta:
I just added for non-existing keys returning `nil`.

solnic:
@esparta thank you, it is merged 😄

esparta:
Nice, thank your for supporting this :)

PR #579, opened 2019-08-14
Fix guarding rules with nested values specified as a hash

### LINKED ISSUE #576: Rule executed despite schema being invalid
Hi :v: Not sure if I use rules correctly, but looks like its a bug. Please take a look. Thank you.

**Describe the bug**

Rule for nested attributes is being executed despite schema contains invalid values.

**To Reproduce**

```ruby
require 'dry/validation'

class ExampleContract < Dry::Validation::Contract
  params do
    required(:contract).schema do
      required(:active_from).value(:date)
      required(:active_until).value(:date)
    end
  end

  rule(contract: [:active_from, :active_until]) do
    key.failure('ops') if value[0] < value[1]
  end
end

f = (Date.today - 1000).to_s
u = (Date.today + 1000).to_s

puts ExampleContract.new.call(contract: { active_from: f, active_until: u }).inspect
#<Dry::Validation::Result{:contract=>{:active_from=>#<Date: 2016-11-02 ((2457695j,0s,0n),+0s,2299161j)>, :active_until=>#<Date: 2022-04-25 ((2459695j,0s,0n),+0s,2299161j)>}} errors={:contract=>{[:active_from, :active_until]=>["ops"]}}>

puts ExampleContract.new.call(contract: { active_from: 'abc', active_until: u }).inspect
# comparison of String with Date failed (ArgumentError)
```

**Expected behavior**

Rule should not be executed when schema is invalid.

solnic:
Fixes #576

### DIFF
--- a/lib/dry/validation/contract.rb
+++ b/lib/dry/validation/contract.rb
@@ -120,7 +120,7 @@ module Dry
         path = Schema::Path[spec]
 
         if path.multi_value?
-          return path.expand.map { |nested_path| error?(result, nested_path) }.any?
+          return path.expand.any? { |nested_path| error?(result, nested_path) }
         end
 
         return true if result.error?(path)
--- a/lib/dry/validation/contract.rb
+++ b/lib/dry/validation/contract.rb
@@ -119,6 +119,10 @@ module Dry
       def error?(result, spec)
         path = Schema::Path[spec]
 
+        if path.multi_value?
+          return path.expand.map { |nested_path| error?(result, nested_path) }.any?
+        end
+
         return true if result.error?(path)
 
         path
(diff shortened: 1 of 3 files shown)


lib/dry/validation/contract.rb
```diff
@@ -119,6 +119,10 @@ def inspect
       def error?(result, spec)
         path = Schema::Path[spec]
 
+        if path.multi_value?
+          return path.expand.map { |nested_path| error?(result, nested_path) }.any?
```
splattael: WDYT about:

```suggestion
          return path.expand.any? { |nested_path| error?(result, nested_path) }
```

Issue #574, opened 2019-07-18
Use existing schema with contract


shanecav84:
I have a schema defined separately. I would like to use it in my contract. Also, it might be worthwhile in the future to be able to extend the existing schema within the contract

## Examples

```ruby
class MySchema < Dry::Schema::Params
  # Schema DSL
end

class MyContract < Dry::Validation::Contract
  params(MySchema) do
    # Extended schema
  end
end
```

## Resources

- Previously mentioned: https://discourse.dry-rb.org/t/how-to-use-existing-schema-with-contract/834

alassek:
I think this is reasonably important to have as a public API, because Contracts cannot be composed together in the same way Schemas can; therefore I need to be able to define them separately.

Currently, I just patch in
```ruby
def self.schema=(value)
  @__schema__ = value
end
```
and then do
```ruby
class MySchema < Dry::Schema::Params
  # Schema DSL
end

class MyContract < Dry::Validation::Contract
  self.schema = MySchema.new
end
```

solnic:
@alassek yes it is, it'll be added in 1.3.0

PR #581, opened 2019-08-15
Don't require block argument with external schema

alassek:
This is a small tweak to the feature added in #580, based on the feature request in #574 

If you are defining a schema externally, it makes sense that you won't always need to extend the schema inside the contract. This PR allows the following:

```ruby
class MySchema < Dry::Schema::Params
  # Schema DSL
end

class MyContract < Dry::Validation::Contract
  params MySchema.new
end
```


solnic:
Thanks Adam. This was a silly oversight on my behalf 😅 I just released 1.3.1 with your fix.

PR #586, opened 2019-09-05
:pencil2: Fixing typos. Grammar

patrickclery:


solnic:
I reckon this is the same as #585

PR #587, opened 2019-09-17
Multiple parent schemas

ianwhite:
This allows dry-validation to make use of the multiple parent feature in dry-schema (https://github.com/dry-rb/dry-schema/pull/191)

This is a BC change, the schema/params/json methods now take one or more parent schemas as an argument


solnic:
Thank you 😄 

> I'm not sure what the procedure is for this.

The procedure is that we'll have a partially broken build until I release the new dry-schema 😅

PR #588, opened 2019-09-17
Bug fix: MessageSet caches empty? too early

ianwhite:
MessageSet in Dry::Validation allows for the set to be manipulated after init.  Its superclass attempts to cache the result of computing #empty? by setting the @empty ivar.

Because of this #empty? will always return true if it is called prior to #add_error.

This pull request fixes that bug by setting @empty to nil when #add_error is called.


solnic:
You're on 🔥 Thank you again!

PR #591, opened 2019-10-04
Replace map flatten with flat_map

grzegorz-jakubiak:
# Replace map flatten with flat_map

Using `flat_map` over `map.flatten` is 1.6 faster according to [Enumerable#map...Array#flatten vs Enumerable#flat_map](https://github.com/JuanitoFatas/fast-ruby#enumerablemaparrayflatten-vs-enumerableflat_map-code).
Failing specs are unrelated i.e. they also fail on `master` branch.

### DIFF
--- a/lib/dry/validation/contract/class_interface.rb
+++ b/lib/dry/validation/contract/class_interface.rb
@@ -162,7 +162,7 @@ module Dry
             .map { |key|
               [key, Schema::Path[key]]
             }
-            .map { |(key, path)|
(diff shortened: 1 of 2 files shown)


solnic:
Thank you 😄 

> Failing specs are unrelated i.e. they also fail on master branch.

Yeah it's unfortunate but the CI only passes with dry-schema from master. It's a temporary situation of course, just need to release new dry-schema 😅

PR #592, opened 2019-10-04
Add an explicit error when failure is configured with an invalid hash

MatElGran:
as discussed here: https://github.com/dry-rb/dry-validation/issues/584


solnic:
Fantastic! Thank you ❤️

PR #595, opened 2019-10-09
Draft PR: Composable contracts 

ianwhite:
Hello! This is a draft PR, in order to get early feedback, as suggested by @solnic.

I've suggested some questions which I think I'd ask, and given my answers.  I'd love to get feedback, of any sort.

### What is it?

An extension for dry-rb that enables contracts to be composed of other contracts.

It looks like this:

```ruby
Dry::Validation.load_extensions(:composable)

class OrderContract < Dry::Validation::Contract
  contract CustomerContract
  path :address do
    contract AddressContract
  end
end
```

### What did you change?

I did some minor refactoring to `contract.rb`, and extracted an interface from `result.rb` (with no functional changes according to the tests) to make my life easier and reduce duplicated code.

I added a bunch of code in `composition.rb` and `composition/*` and hooked that code into Contract via the extension mechanism.

### What problem does it solve?

For me, it solves two problems.

1. reusing contracts, including their rules
2. applying contracts to nested hashes of the input, without having to write the rules such that they know about their position in the tree.

### What does it not do yet?

Current limitations that I can see are

- no support for sharing context between contracts
- no way of specifying a contract for each element of **array** input
- no way of conditionally applying a contract to the input
- specific interaction with other extensions is not well tested

### Anything else?

Nope, thanks for your time!

### DIFF
--- a/lib/dry/validation.rb
+++ b/lib/dry/validation.rb
@@ -27,6 +27,10 @@ module Dry
       require 'dry/validation/extensions/predicates_as_macros'
     end
 
+    register_extension(:composable) do
+      require 'dry/validation/extensions/composable'
+    end
+
     # Define a contract and build its instance
     #
     # @example
--- a/lib/dry/validation/composition.rb
+++ b/lib/dry/validation/composition.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+require 'dry/validation/composition/step'
+
+module Dry
+  module Validation
+    # A Composition is a list of steps
+    #
+    # @see Dry::Validation::Composable
+    #
+    # @api private
+    class Composition
+      include Dry::Equalizer(:steps, inspect: false)
+
+      # @return [Array]
+      #
+      # @api private
+      attr_reader :steps
+
+      # @param [Array] initial steps array
+      def initialize(steps = EMPTY_ARRAY)
+        @steps = steps.dup
+      end
+
+      def inspect
+        steps_str = steps.map do |s|
+          s.prefix ? "#{s.prefix.to_a.join(DOT)} => #{s.contract}" : s.contract
+        end.join(', ')
+
+        "#<#{self.class.name} steps=[#{steps_str}]>"
+      end
+
+      # @return [Bool]
+      #
+      # @api private
+      def empty?
+        steps.empty?
+      end
+
+      # apply our steps to the input recording in result
+      #
+      # @param [Composition::Result]
+      # @param [Hash]
+      #
+      # @return [Composition::Result]
+      #
+      # @api private
+      def call(input, result = Composition::Result.new)
+        steps.reduce(result) { |final_result, step| step.call(input, final_result) }
+      end
+
+      # add a new step to the composition
+      #
+      # @param [Contract]
+      # @param [Schema::Path] optional
+      #
+      # @api private
+      def add_step(contract, prefix)
+        steps << Step.new(contract, prefix)
+        self
+      end
+    end
+  end
+end
--- a/lib/dry/validation/composition/builder.rb
+++ b/lib/dry/validation/composition/builder.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+require 'dry/schema/path'
+
+module Dry
+  module Validation
+    class Composition
+      # Adds steps to a composition via #contract and #path methods
+      #
+      # @api private
+      class Builder
+        # @return [Composition]
+        #
+        # @api private
+        attr_reader :composition
+
+        # @return [Schema::Path]
+        #
+        # @api private
+        attr_reader :prefix
+
+        def initialize(composition, prefix = nil)
+          @composition = composition
+          @prefix = Schema::Path[prefix] if prefix
+        end
+
+        def contract(contract, path: nil)
+          composition.add_step contract, prefixed(path)
+        end
+
+        def path(path, &block)
+          self.class.new(composition, prefixed(path)).instance_eval(&block)
+        end
+
+        private
+
+        def prefixed(path)
+          path = Schema::Path[path] if path
+          path = Schema::Path.new([*prefix, *path]) if prefix
+          path
+        end
+      end
+    end
+  end
+end
--- a/lib/dry/validation/composition/result.rb
+++ b/lib/dry/validation/composition/result.rb
@@ -1,0 +1,160 @@
+# frozen_string_literal: true
+
+require 'dry/validation/result_interface'
+
+module Dry
+  module Validation
+    class Composition
+      # A Composition::Result is built from many results with optional
+      # input prefixes.
+      #
+      # @see ResultInterface
+      #
+      # @api public
+      class Result
+        include ResultInterface
+
+        include Dry::Equalizer(:errors, :values, inspect: false)
+
+        # Build a new result
+        #
+        # @param [Dry::Schema::Result] schema_result
+        #
+        # @api private
+        def self.new(options = EMPTY_HASH)
+          result = super
+          yield(result) if block_given?
+          result.freeze
+        end
+
+        # Result options
+        #
+        # @return [Hash]
+        #
+        # @api private
+        attr_reader :options
+
+        # Result values
+        #
+        # @return [Values]
+        #
+        # @api public
+        attr_reader :values
+
+        # prefixed results
+        #
+        # @return [Array[Schema::Path, Result]]
+        #
+        # @api private
+        attr_reader :prefixed_results
+
+        # Initialize a new result
+        #
+        # @api private
+        def initialize(options)
+          @options = options
+          @values  = Values.new({})
+          @errors  = MessageSet.new([], options)
+
+          # prefixed results is only necessary so that we can rebuild error
+          # messages when passed options to #errors.
+          #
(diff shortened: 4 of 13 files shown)


ianwhite:
(I just thought I'd mention that I will clean up all the code climate stuff, and squash into a new branch, if people decide that this is a good way to go - sorry for all that noise)

waiting-for-dev:
This feature would be very cool @ianwhite In general, I think that we should extend this way of thinking to other dry-rb gems and rely more on composition rather than inheritance (or at least support both).

I guess you can push new commits to this branch so that we don't clutter your work.

I think that we should make progressive steps with this feature. I think first thing to do should be adding schema composition to dry-schema gem. Then, dry-validation could use it and add rules composition on top.

About the `path/contract` tuple. Is it not enough the already present feature of reusing schemes?

```ruby
required(:address).hash(AddressSchema)
```

solnic:
> I think that we should extend this way of thinking to other dry-rb gems and rely more on composition rather than inheritance (or at least support both).

Yep. All dry-rb libraries are very well suited for composition, because in most cases we're dealing with functional processing. That's why in case of this PR it should be easier to apply multiple contracts and merge results rather than have a complex inheritance hierarchy.

> But I think we are mixing two concerns here...
> Composing contracts
> Reusing contracts in a nested key (which may or not compose other contracts)

That is a good point. Splitting this work into two parts would make sense. Would that be OK for you @ianwhite ?

> Allow .hash to take a contract: required(:address).hash(AddressContract).

This is not a good idea. I'd prefer if dry-schema didn't have to handle more argument types. Nesting schemas is one of the most complex parts of this lib. I think dry-validation could extract a schema from another contract and pass it to `hash` method, then deal with merging rules separately.

waiting-for-dev:
> This is not a good idea. I'd prefer if dry-schema didn't have to handle more argument types. Nesting schemas is one of the most complex parts of this lib. I think dry-validation could extract a schema from another contract and pass it to hash method, then deal with merging rules separately.

Ok, I haven't studied carefully this part of the code. What I had in mind was pretty simple. I guess in some part is expected that `#hash` can accept a `Dry::Schema::Processor`. Instead of that we could check for anything responding to `#schema`. In the case of a `Dry::Schema::Processor` the implementation would return `self`. For a `Dry::Validation::Contract` it already returns the associated schema, if I'm not wrong. I just wanted to describe it a little bit just in case my previous description sounded as something more complicated.

ianwhite:
Thanks for the comments @solnic & @waiting-for-dev !

> > 1) Composing contracts
> > 2) Reusing contracts in a nested key (which may or not compose other contracts)
>
> That is a good point. Splitting this work into two parts would make sense. Would that be OK for you @ianwhite ?

After thinking about this for a bit, I agree!  I'd like to lay out my understanding of what is required.  Please correct/comment.

## TLDR;

I think we need 2 new features, I'm happy to work on them.

1. composbale schemas/contracts via new `#compose` method
2. mounting contract (not schemas) behaviour at paths via new `#at_path` method

## In detail

@waiting-for-dev identified 2 features:

1. `self` applies behaviour from `other` to the input at root
2. `self` applies behaviour of `other` to the input at `path`

### My understanding of current behaviour

`Dry::Schema` currently allows (1) by inheritance, and allows (2) by its API.

`Dry::Validation::Contract` gets the bevahiour of Dry::Schema for free for schemas only.  Otherwise it allows (1) by inheritance, and (2) is currently not possible.

### Proposed features

#### 1. Composition

This feature is to implement (1) via a composition mechanism.

This feature should apply to both `Dry::Schema` and `Dry::Validation::Contract`.  Given this, it's interface method should be **`#compose`** as per @waiting-for-dev's suggestion.

```ruby
# setup
NameSchema = Dry::Schema.define { required(:name).filled(:string) }
EmailSchema = Dry::Schema.define { required(:email).filled(:string) }
StartsAtContract = Dry::Validation.Contract do
  schema { required(:starts_at).filled(:date) }
  rule(:starts_at) { key.failure('is in the past') if value < Date.now } 
end

# composed schema
ComposedSchema = Dry::Schema.define do
  compose EmailSchema, NameSchema
end

# composed contract
class ComposedContract < Dry::Validation::Contract
  compose NameSchema, EmailSchema, StartsAtContract
end
```

#### 2. Composition at path

This feature is not required for `Dry::Schema` as it already has it, so it only applies to contracts.

The motivation for this is so we can write rules without knowledge of their position in the input tree.

Because of this, I think that this feature should apply all of the contract ClassInterface.  I propose that the method be called **`#at_path`**.

```ruby
# composed contract
class ComposedContract < Dry::Validation::Contract
  compose NameSchema, EmailSchema, StartsAtContract

  at_path :address do
    compose AddressContract

    # but could also do.
    # rule ...
    # schema ...
  end
end
```

### Closing

Summary: feature 1) is composbale schemas/contracts via #compose method.  feature 2) is mounting contract behaviour at paths via #at_path method. 

As @waiting-for-dev says, we should either coerce or raise an error when contracts/schemas are composed when they are of different types (json/params/etc).  What should we do?

My use case also needs the ability to conditionally apply contracts to the input.  In prior discussion with @solnic, we agreed to put that on the backburner.  I'd like to just signpost this is case you have thoughts about it which should be take into consideration early.  I'm thinking of a rule like DSL, but its job is not to add errors, but to decide what contracts to apply.

(as a new contributor to dry-rb, I am very happy to be educated re: any etiquette breaches/norms not covered in the code of conduct)

ianwhite:
Actually `Dry::Schema` allows composition via the `:parent` option (now with multiple parents).  I think we could add a `#compose` method based on that if that's desirable.

So I guess there are 3 bits of work, in order.

1. `Dry::Schema`: add `#compose`, and document it
2. `Dry::Validation`: add `#compose`, and document it
3. `Dry::Validation`: add some method of composing at a path (syntax/semantics TBD)

I'm happy to work on these, once I get a green light.

waiting-for-dev:
```ruby
> # composed contract
> class ComposedContract < Dry::Validation::Contract
>   compose NameSchema, EmailSchema, StartsAtContract
> end# composed contract
```

I see here `compose` can take both schemes and contracts as arguments. While I'm not 100% against it, I don't like it because it implies implementation must ask what the argument is (`is_a?`).

I'd would prefer for `compose` to only take contracts as arguments. What I think it would make sense is to have another method `compose_schema` which would delegate to dry-schema's `compose` method. I think of dry-validation's `compose` implementation as internally calling `compose_schema` and another method `compose_rules`. However, probably it makes no sense to expose `compose_rules` as a public interface, as I can't imagine a use case where you want to compose rules from a contract but not its schema.

```ruby
> # composed contract
> class ComposedContract < Dry::Validation::Contract
>   compose NameSchema, EmailSchema, StartsAtContract
>
>   at_path :address do
>    compose AddressContract
>
>    # but could also do.
>    # rule ...
>    # schema ...
>  end
> end
```

Yep, but to be consistent with:

```ruby
required(:address).hash(AddressSchema)
```

we should be able to do:

```ruby
at_path(:address, AddressContract)
```

What I said before of being able to do:

```ruby
rule(:address, AddressContract)
```

surely it makes no sense (again, who wants to use rules from an external contract but not its schema?).

> As @waiting-for-dev says, we should either coerce or raise an error when contracts/schemas are composed when they are of different types (json/params/etc). What should we do?

My view is that we should raise. Not straightforward to decide what the user was trying to do when mixing schemes of different type...

> Actually Dry::Schema allows composition via the :parent option (now with multiple parents). I think we could add a #compose method based on that if that's desirable.

Good point. It does the same but using OO nomenclature...

ianwhite:
> I see here compose can take both schemes and contracts as arguments. While I'm not 100% against it, I don't like it because it implies implementation must ask what the argument is (is_a?).

Yep, agreed. 👍 (also re: compose_rules internal method)

> Yep, but to be consistent with: `required(:address).hash(AddressSchema)`...

The reason I'd like some sort of block syntax here is that I'd like to compose a contract using smaller contracts, and be able to have the knowledge the overall input data structure in just the composing contract.  I think we could combine syntax to something like:

```ruby
at_path :addresses do
  at_path(:billing, AddressContract)
  at_path(:delivery, AddressContract)
end
```

But I think we can move forward without having decided on this, as it's the last of the 3 pieces of work that needs to be done?
  
> My view is that we should raise. Not straightforward to decide what the user was trying to do when mixing schemes of different type...

Agreed 👍 

So, in summary, the work is to:

1. Dry::Schema: add #compose([Schema]), and document it
2. Dry::Validation: add #compose([Contract]), and #compose_schema([Schema]), and document it
3. Dry::Validation: add some method of composing at a path (syntax/semantics TBD)

I'll wait for @solnic to chime in before moving forward.  Thanks @waiting-for-dev!

waiting-for-dev:
> The reason I'd like some sort of block syntax here is that I'd like to compose a contract using smaller contracts, and be able to have the knowledge the overall input data structure in just the composing contract.

Yeah, what I meat was to support both: a contract argument or a block which leaves you in a contract definition context. Just like `#hash` with schemes :)

solnic:
> I'll wait for @solnic to chime in before moving forward. 

The plan makes sense. You could add new APIs to dry-schema in this repo and once everything is ready we could extract it (if you want). From my experience this is a faster way of working.

PR #596, opened 2019-10-11
[docsite] fix links

skryukov:
Also should be merged in master =)


solnic:
Thanks :) I'll automate that eventually (somehow).

PR #605, opened 2019-11-23
Update index.html.md

wintersolutions:
Fixed a small discrepancy between example message "must be greater than 18" and expression `if value < 18`


skryukov:
Thanks @wintersolutions =)

PR #613, opened 2020-01-08
Add support for pattern matching on result values

flash-gordon:
Two possible ways of matching a result:

1. With keys
```ruby
case contract_result
in foo:, bar: 'baz'
  # same as in dry-schema
end
```

2. With arrays (tuple, pair of hash values)
```ruby
case contract_result
in { value_from_input: }, { context_value: }
  # rules can add values to context that after can be destructured like above
end
```


flash-gordon:
thanks @dry-bot

Issue #608, opened 2019-12-19
import_predicates_as_macros: included_in?, format?, size? issues


aleksandra-stolyar:
```ruby
class MyContract < Dry::Validation::Contract
  params do
    required(:hogwarts_house).hash do
      required(:head).filled(:string)
      required(:common_room).hash do
        required(:name).filled(:string)
        required(:location).filled(:string)
      end
    end
  end

  rule('hogwarts_house.head').validate(size?: 1..255)
  rule('hogwarts_house.head').validate(format?: SOME_REGEX)
  rule('hogwarts_house.common_room.location').validate(included_in?: ['tower', 'underground'])
end
```

For `included_in?` and `size?` passing an array of integers/strings not working but it works when just one value passed or with predicates which do not require arrays)
`included_in?`
> `rule('hogwarts_house.common_room.location').validate(included_in?: ['tower', 'underground'])`

`size?`
>`rule('hogwarts_house.head').validate(size?: 1..255)`
ArgumentError (wrong number of arguments (given 256, expected 2))

This predicate is just returning error.
`format?`
>`rule('hogwarts_house.head').validate(format?: SOME_REGEX)`
<Dry::Container::Error: Nothing registered with the key :format?>

lromeo:
Just ran into these same issues with `included_in?` and `format?`

waiting-for-dev:
#610 adds missing `:format?` predicate.

About the other two, the issue is that they are related to predicates expecting an array (or range) as argument. It hasn't straightforward solution. Macro argument(s) are forwarded to the predicate method, and we don't have any way to tell whether it expects an array or an unitary item. I see three options:

- Force the user to be explicit. For example, currently the following works (@aleksandra-stolyar @lromeo take note for now):

```ruby
rule('hogwarts_house.head').validate(size?: [1..255])
rule('hogwarts_house.common_room.location').validate(included_in?: [['tower', 'underground']])
```

- Manually keep record of what predicates expect what.

- Modify `dry-logic` so that the predicates with array or range arguments also work with splat arguments.

solnic:
@waiting-for-dev that's OK, it's **always worked like that**, going back to dry-validation `0.x`. We gotta make sure to document it properly.

solnic:
@waiting-for-dev  ...no wait, sorry - it did not work like that in dry-validation `0.x`! dooh. OK this needs to be investigated because I can't say why it's not working as expected in case of predicate extension :(

[...]

PR #616, opened 2020-01-15
Fix macros expecting a range as option

waiting-for-dev:
References #608

### DIFF
--- a/lib/dry/validation/rule.rb
+++ b/lib/dry/validation/rule.rb
@@ -118,10 +118,16 @@ module Dry
         args.each_with_object([]) do |spec, macros|
           case spec
           when Hash
-            spec.each { |k, v| macros << [k, Array(v)] }
+            add_macro_from_hash(macros, spec)
           else
             macros << Array(spec)
           end
+        end
+      end
+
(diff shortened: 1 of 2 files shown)


lib/dry/validation/rule.rb
```diff
@@ -118,12 +118,21 @@ def parse_macros(*args)
...
           end
         end
       end
+
+      def add_macro_from_hash(macros, spec)
+        spec.each do |k, v|
+          macros << [
+            k,
+            v.is_a?(Array) ? v : [v]
+          ]
```
solnic: could you inline this?

waiting-for-dev:
Do you mean the whole `each` block or just the array push? In the first
case, maybe line length is too high...

solnic:
@waiting-for-dev just the array push. re line-length, I think these days with gigantic screen resolutions `100` is a sensible value, but we can change it to more standard 80

waiting-for-dev:
Ok @solnic , I forced a push with the change style. About line length, I'm always nostalgic of the old times, so I like the 80 limit, even if I recognize it's sometimes insane :smile:

Issue #615, opened 2020-01-15
Add key.success? and key.failure? predicates


flash-gordon:
In complex rules, it may be useful to know whether the key has any failure message already.

```ruby
rule(:foos).each do
  key.success? # => true
  key.failure? # => false
  key.failure('wrong')
  key.success? # => false
  key.failure? # => true
end
```

waiting-for-dev:
If it's ok, I'll give a try to it.

flash-gordon:
Sure.

waiting-for-dev:
Hmmm this seems tricky... it implies changing the architecture so that the evaluator and failures for every key are shared between all rules. Am I wrong?

flash-gordon:
For me, it would be enough to work only within a single rule run. Though I can see how it can be not immediately obvious. @solnic any thoughts?

flash-gordon:
`rule.failure?` or something along the lines could be another approach

waiting-for-dev:
Probably `rule.failure?` makes more sense, as `key.failure?` implies that `key(name).failure?` should work. However, the scenario where evaluator is shared also makes sense for even more complex rules :)

solnic:
I forgot we have `error?`. Currently it expects path arg but we can set it to the default key, so you could have:

```ruby
rule(:foo) do
  error? # expands to `error?("foo")`
  
  error?("other.path") # if you need to check something else
end
```

I think this should be enough and we don't need `key.failure/success?`.

flash-gordon:
works for me though `rule.error?`/`rule.failure?` still could be useful

solnic:
@flash-gordon why?

flash-gordon:
@solnic rules can be quite convoluted, especially in macros. This is the exact need I had when filed this issue. Keeping them simple is not always feasible.

solnic:
@flash-gordon ok makes sense

flash-gordon:
@waiting-for-dev so I think adding a default argument to `error?` would be enough here. I can add a separate feature request about `rule.error?`.

PR #617, opened 2020-01-20
Add `Evaluator#schema_error?` and `Evaluator#rule_error?`

waiting-for-dev:
This tries to implement what has been discussed in #615 

@flash-gordon @solnic please, take a look at the WIP code... the spec is failing because `#error?(:email)` returns false, even when it has an error added in the first rule. It's because `Evaluator#error?` delegates to `Result#error?`, which in turn delegates to `Dry::Schema::Result#error?`. Is it intended behaviour? I mean, we are looking just to the errors set in the schema phase and ignoring rules.

### DIFF
--- a/docsite/source/rules.html.md
+++ b/docsite/source/rules.html.md
@@ -203,8 +203,7 @@ PersonContract.new.(email: nil, name: 'foo').errors.to_h
 # { email: ['must be a string'], name: ['first introduce a valid email'] }
 ```
 
-In complex rules you may be interested to know whether the current rule already
-had an error. For that, you can use `#rule_error?`
+In complex rules you may be interested to know whether the current rule already had an error. For that, you can use `#rule_error?`
 
 ```ruby
 class FooContract < Dry::Validation::Contract
--- a/docsite/source/rules.html.md
+++ b/docsite/source/rules.html.md
@@ -181,6 +181,47 @@ contract.call(email: 'jane@doe.org', login: 'jane', password: "").errors.to_h
 # => {:password=>["password is required"]}
 ```
 
+### Checking for previous errors
+
+Sometimes you may be interested in adding an error when some other error has happened.
+
+With `#schema_error?(key)` you can check whether the schema has an error for a given key:
+
+```ruby
+class PersonContract < Dry::Validation::Contract
+  schema do
+    required(:email).filled(:string)
+    required(:name).filled(:string)
+  end
+
+  rule(:name) do
+    key.failure('first introduce a valid email') if schema_error?(:email)
+  end
+end
+
+PersonContract.new.(email: nil, name: 'foo').errors.to_h
+# { email: ['must be a string'], name: ['first introduce a valid email'] }
+```
+
+In complex rules you may be interested to know whether the current rule already
+had an error. For that, you can use `#rule_error?`
+
+```ruby
+class FooContract < Dry::Validation::Contract
+  schema do
+    required(:foo).filled(:string)
+  end
+
+  rule(:foo) do
+    key.failure('failure added')
+    key.failure('failure added after checking') if rule_error?
+  end
+end
+
+FooContract.new.(foo: 'foo').errors.to_h
+# { foo: ['failure added', 'failure added after checking'] }
+```
+
 ### Defining a rule for each element of an array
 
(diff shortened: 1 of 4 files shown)


flash-gordon:
@solnic then `rule_error?` should only look up rule errors but it's not clear whether it should a single-rule check or all-rules check. I would say the name suggests it's current-rule-only.

waiting-for-dev:
I think that current name is misleading... what do you think about adding for now `#schema_error?` and `#rule_error?`, and adding a deprecation warning for `#error?`? I think the latest should perform both checks in a future dry-schema release...

waiting-for-dev:
In order to keep a thin interface another option is to overload `#error?` method to accept a keyword argument `from:`, expecting it to take `:schema`, `:rule` or `:all` as values. To keep backward compatibility it should default to `:schema`, but I'd change it to `:all` in 2.0

solnic:
@waiting-for-dev I was going to suggest `rule_error?` + `schema_error?` too. I think it's fine that `rule_error?` would check all-the-rules. We can always extend it later on to be able to target a specific rule

flash-gordon:
I'm for it

waiting-for-dev:
Hmm not sure I'm following here... Currently, `#error?` accepts a path as argument, so it's already capable of targeting a single key. Should `#scheme_error?` be an alias of `#error?`. It'd make sense to me. Or are you talking about checking all keys at once? The same applies to `#rule_error?`. Why not accept a path as argument from the beginning?

solnic:
> Should #schema_error? be an alias of #error?

Yes, and `#error?` should be deprecated and removed in 2.0.0.

> The same applies to #rule_error?. Why not accept a path as argument from the beginning?

My understanding is that we want to check if a rule failed, doesn't matter under which key, because a rule can fail due to multiple reasons.

@flash-gordon have you considered that ☝️ ?

flash-gordon:
I don't think having `path` argument for `rule_error?` would be particularly useful. This is based on my experience. If you want to check if other rules failed you likely want to do it explicitly. For that, we need a means of nesting rules or something like this. If you want to check failures in the current rule then it's more likely you don't care about paths. This is what I learned from my experience. The main argument for my reasoning is adding new rules should not affect existing ones.

waiting-for-dev:
Hey @solnic @flash-gordon you can check it now

lib/dry/validation/evaluator.rb
```diff
@@ -163,17 +167,26 @@ def key?
...
       end
 
+      # Check if there are any errors on the current rule
+      #
+      # @return [Boolean]
+      #
+      # @api public
+      def rule_error?
+        key(path).opts.any?
+      end
```
solnic: Could you add `Failures#empty?` and use it here instead of inspecting its state?

waiting-for-dev: Here it goes @solnic

lib/dry/validation/failures.rb
```diff
@@ -60,6 +60,11 @@ def failure(message, tokens = EMPTY_HASH)
         opts << { message: message, tokens: tokens, path: path }
         self
       end
+
+      # @api private
+      def empty?
+        opts.any?
```
solnic: this should be `opts.empty?` and in `rule_error?` it should do `!key(path).empty?`

waiting-for-dev: Oh, damn it! I think yesterday I had my brain upside down... Now it's ok :sweat_smile:

docsite/source/rules.html.md
```diff
@@ -181,6 +181,47 @@ contract.call(email: 'jane@doe.org', login: 'jane', password: "").errors.to_h
...
+    key.failure('first introduce a valid email') if schema_error?(:email)
+  end
+end
+
+PersonContract.new.(email: nil, name: 'foo').errors.to_h
+# { email: ['must be a string'], name: ['first introduce a valid email'] }
+```
+
+In complex rules you may be interested to know whether the current rule already
+had an error. For that, you can use `#rule_error?`
```
solnic: Please don't use hard-wrapping in docs, it can actually screw up MD formatting AND it makes writing docs really annoying

waiting-for-dev: Apologies... I fixed it :sweat_smile:

flash-gordon:
Thank you @waiting-for-dev!

PR #633, opened 2020-03-15
doc: YARD annotations with more documentation

olleolleolle:
-  an attempt to reduce warnings and still keep the documentation legible

## Background

Before this change, there were a few warnings when generating documentation.

> Building YARD (yri) index for dry-validation-1.5.0...
lib/dry/validation/macros.rb:28: [UnknownParam] @param tag has unknown parameter name: *args. Did you mean `args`?
lib/dry/validation/values.rb:38: [UnknownParam] @param tag has unknown parameter name: key
lib/dry/validation/evaluator.rb:149: [UnknownTag] Unknown tag @public
lib/dry/validation/evaluator.rb:172: [UnknownParam] @param tag has unknown parameter name: A

## Questions for reviewer

- Are there technical terms for the Schema::Path-compatible argument I tried to document, which was called "key"?

### DIFF
--- a/lib/dry/validation/evaluator.rb
+++ b/lib/dry/validation/evaluator.rb
@@ -146,7 +146,7 @@ module Dry
       #
       # @return [Object]
       #
-      # @public
+      # @api public
       def value
         values[key_name]
       end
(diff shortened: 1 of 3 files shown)


solnic:
> Are there technical terms for the Schema::Path-compatible argument I tried to document, which was called "key"?

@olleolleolle I plan to make this fully compatible with the json path spec but we're not there yet so let's maybe stick to "path"

PR #636, opened 2020-04-06
Don't require dry-monads in hints extension.

schokomarie:
It is not needed here and leads to an error when not using dry-monads


flash-gordon:
OMG, thanks!

Issue #655, opened 2020-07-17
Dry::Validation::Result#error?(:field) returns false although there is an error (when triggered by rule)


PragTob:
Hello my friends :wave: : 

Thanks for all your hard work on these gems :green_heart: 

**Describe the bug**

When an error is the result of a rule it seems that `error?(:field)` returns false, although there is an error.

**To Reproduce**

```ruby
require 'dry-validation'

class Nameless < Dry::Validation::Contract
  schema do
    required(:name).filled(:string)
  end

  rule(:name) do
    key.failure('womp womp womp')
  end
end

# change me to empty string/nil to see that error? works in general (for schema)
result = Nameless.new.call(name: 'ABC')

p result # #<Dry::Validation::Result{:name=>"ABC"} errors={:name=>["womp womp womp"]}>

puts "failure? #{result.failure?}" # failure? true
puts "errors: #{result.errors.inspect}" # errors: #<Dry::Validation::MessageSet messages=[#<Dry::Validation::Message text="womp womp womp" path=[:name] meta={}>] options={}>
puts "errors.to_h: #{result.errors.to_h}" # errors.to_h: {:name=>["womp womp womp"]}
puts "result.error?(:name): #{result.error?(:name)}" # result.error?(:name): false
```

**Expected behavior**

`error?` should return `true` because there is an error.

I'm semi worried that because I found `error?` in the `dry-schema` docs that it only works for schema errors but it'd be greatly misleading then. I'd propose to make it work. Failing at that, I'd remove the method from `Validation::Result` as it'd be highly misleading.

**Your environment**

- Affects my production application: kinda (I'm upgrading lots of gems and instead of changing my tests to be `.errors.to_h` and check with `include` using `error?(:field)` seemed like a great way.
- dry-gems versions: all up to date (dry-validation 1.5.2)

<details>

```
bash-5.0# grep 'dry-' Gemfile.lock
    dry-configurable (0.11.6)
      dry-core (~> 0.4, >= 0.4.7)
      dry-equalizer (~> 0.2)
    dry-container (0.7.2)
      dry-configurable (~> 0.1, >= 0.1.3)
    dry-core (0.4.9)
    dry-equalizer (0.3.0)
    dry-inflector (0.2.0)
    dry-initializer (3.0.3)
    dry-logic (1.0.6)
      dry-core (~> 0.2)
      dry-equalizer (~> 0.2)
    dry-schema (1.5.2)
[...]

PR #656, opened 2020-07-17
Make Result#error?(key) return true for rule and schema errors

PragTob:
Not super happy with this as of now :) Exchanged usage of
error? with schema_error? in the code base as otherwise there
thankfully were a bunch of failures.

The implementation of error? itself is basically stolen from
its counterpart at `Schema::Result`. It makes sense to check
if there are any errors for this but not sure if we shouldn't
rather still call out to schema_result.error?

Also, root/base behavior is a bit of a mystery.

Fix attempt for #655

### DIFF
--- a/lib/dry/validation/result.rb
+++ b/lib/dry/validation/result.rb
@@ -101,8 +101,15 @@ module Dry
 
       # Check if values include an error for the provided key
       #
-      # @api private
+      # @api public
       def error?(key)
+        errors.any? { |msg| Schema::Path[msg.path].include?(Schema::Path[key]) }
+      end
+
+      # Check if the base schema (without rules) includes an error for the provided key
+      #
+      # @api private
+      def schema_error?(key)
         schema_result.error?(key)
       end
 
--- a/spec/integration/result_spec.rb
+++ b/spec/integration/result_spec.rb
@@ -50,6 +50,33 @@ RSpec.describe Dry::Validation::Result do
     end
   end
 
+  describe "#error?" do
+    subject { result }
+
+    let(:schema_result) do
+      double(:schema_result, message_set: [], to_h: {email: "jane@doe.org"})
+    end
+
+    let(:result) do
+      Dry::Validation::Result.new(schema_result) do |r|
+        r.add_error(Dry::Validation::Message.new("root error", path: [nil]))
+        r.add_error(Dry::Validation::Message.new("email error", path: [:email]))
+      end
+    end
+
+    it "reports an error on email" do
+      expect(subject.error?(:email)).to eq true
+    end
+
+    it "reports an error on 'root' when asked for [nil]" do
+      expect(subject.error?([nil])).to eq true
+    end
+
+    it "doesn't report errors on non existing keys" do
+      expect(subject.error?(:nonexistant)).to eq false
+    end
+  end
+
   describe "#inspect" do
     let(:params) do
       double(:params, message_set: [], to_h: {})
--- a/spec/integration/result_spec.rb
+++ b/spec/integration/result_spec.rb
@@ -51,8 +51,6 @@ RSpec.describe Dry::Validation::Result do
   end
 
   describe "#error?" do
-    subject { result }
-
     let(:schema_result) do
       double(:schema_result, message_set: [], to_h: {email: "jane@doe.org"})
     end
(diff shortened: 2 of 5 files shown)


lib/dry/validation/result.rb
```diff
@@ -101,8 +101,16 @@ def failure?
 
       # Check if values include an error for the provided key
       #
-      # @api private
+      # @api public
```
PragTob, line 104: not sure if that's alright but imo would be great to have it for public consumption :)

spec/integration/result_spec.rb
```diff
@@ -50,6 +50,33 @@
...
+        r.add_error(Dry::Validation::Message.new("email error", path: [:email]))
+      end
+    end
+
+    it "reports an error on email" do
+      expect(subject.error?(:email)).to eq true
+    end
+
+    it "reports an error on 'root' when asked for [nil]" do
+      expect(subject.error?([nil])).to eq true
```
PragTob, line 72: I added this because it's current behavior (given the root error path that I just copied from the other test). Not sure if it should be like that or if there's a nicer way?

spec/integration/result_spec.rb
```diff
@@ -50,6 +50,33 @@
...
+
+    let(:schema_result) do
+      double(:schema_result, message_set: [], to_h: {email: "jane@doe.org"})
+    end
+
+    let(:result) do
+      Dry::Validation::Result.new(schema_result) do |r|
+        r.add_error(Dry::Validation::Message.new("root error", path: [nil]))
+        r.add_error(Dry::Validation::Message.new("email error", path: [:email]))
+      end
```
PragTob, line 64: this is basically copy & past from above. Not sure what feelings are around here for contexts and distance to `let`s :)

spec/integration/result_spec.rb
```diff
@@ -50,6 +50,33 @@
...
+    end
+
+    it "reports an error on email" do
+      expect(subject.error?(:email)).to eq true
```
solnic, line 68:
```suggestion
      expect(subject.error?(:email)).to be(true)
```

spec/integration/result_spec.rb
solnic, line 72:
```suggestion
      expect(subject.error?([nil])).to be(true)
```

spec/integration/result_spec.rb
```diff
@@ -50,6 +50,33 @@
     end
   end
 
+  describe "#error?" do
+    subject { result }
```
solnic, line 54: No need for this, please just use `result` in the examples

spec/integration/result_spec.rb
```diff
@@ -50,6 +50,33 @@
...
+    it "reports an error on email" do
+      expect(subject.error?(:email)).to eq true
+    end
+
+    it "reports an error on 'root' when asked for [nil]" do
+      expect(subject.error?([nil])).to eq true
+    end
+
+    it "doesn't report errors on non existing keys" do
+      expect(subject.error?(:nonexistant)).to eq false
```
solnic, line 76:
```suggestion
      expect(subject.error?(:nonexistant)).to be(false)
```

spec/integration/result_spec.rb
solnic: Yes base errors are represented by `nil` key, doing `error?(nil)` would work too.

lib/dry/validation/result.rb
solnic: Yes with this PR we can make it a public API. It was my intention from the beginning, actually.

spec/integration/result_spec.rb
PragTob: I double checked and error?(nil) doesn't work.

```
  1) Dry::Validation::Result#error? reports an error on 'root' when asked for nil
     Failure/Error: errors.any? { |msg| Schema::Path[msg.path].include?(Schema::Path[key]) }
     
     ArgumentError:
       +spec+ must be either a Symbol, Array, Hash or a Path
     # ./lib/dry/validation/result.rb:106:in `block in error?'
     # ./lib/dry/validation/result.rb:106:in `any?'
     # ./lib/dry/validation/result.rb:106:in `error?'
     # ./spec/integration/result_spec.rb:70:in `block (3 levels) in <top (required)>'

Finished in 0.61031 seconds (files took 0.60271 seconds to load)
184 examples, 1 failure

Failed examples:

rspec ./spec/integration/result_spec.rb:69 # Dry::Validation::Result#error? reports an error on 'root' when asked for nil
```

Should it work? Then I'll see how it could be made to work but it seems like that might be in `Schema::Path[key]` then aka in dry-schema :)

PragTob:
PR updated :tada: Although there's a question about `error?(nil)` which doesn't work atm. See further above :)

spec/integration/result_spec.rb
solnic: @PragTob ah right, I forgot that `Path` doesn't turn `nil` into `[nil]` automatically, so yes, this is expected. BTW - for base errors (aka root errors), we have `Result#base_error?` method.

PR #660, opened 2020-07-27
Fix base error predicate

### LINKED ISSUE #659: Rule#each raise an error if the schema is extended by another one
Hello Guys,

First of all, thank you for your gems which are really good :)

**Describe the bug**

An error is raised when we use the `Rule#each` method for a key when another key, **defined before the array**, does not pass the schema validation :

```bash
NoMethodError: undefined method `each_with_index' for "wrong_type":String
Did you mean?  each_line
```

The error seem to be raised from here : https://github.com/dry-rb/dry-validation/blob/master/lib/dry/validation/rule.rb#L86

**To Reproduce**

```ruby
require 'dry-validation'

class TestContract < Dry::Validation::Contract
  params do
    required(:something).filled(:string)
    required(:my_array).array(:hash) do
      optional(:test).filled(:string)
    end
  end

  rule(:my_array).each do

  end
end

# `something: nil` doesnt pass the schema validation
TestContract.new.(something: nil, my_array: 'wrong_type').errors.to_h
```

**Expected behavior**

It `.each` block or the macros should not be executed when the schema is not correct.

**Your environment**

- Affects my production application: **YES**
- Ruby version: 2.7.0
- OS: MacOS

solnic:
Fixes #659

### DIFF
--- a/lib/dry/validation/result.rb
+++ b/lib/dry/validation/result.rb
@@ -123,7 +123,7 @@ module Dry
           key_path = Schema::Path[key]
           err_path = Schema::Path[error.path]
 
-          return false unless key_path.same_root?(err_path)
+          next unless key_path.same_root?(err_path)
 
           key_path == err_path
         }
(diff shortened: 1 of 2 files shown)


flash-gordon:
I do not like `return`s

PR #662, opened 2020-07-30
Fix full error messages

### LINKED ISSUE #661: Full error messages don't include the key name
_This issue was initially reported in discourse: https://discourse.dry-rb.org/t/inconsistency-in-resolving-full-error-messages/1091_

1. Full message of a key failure does not include the key name when the key is not translated and the failure method was called with a symbol message, e.g. `key.failure(:is_invalid)`. 
So the error message looks like `" is invalid"`, instead of `"foo is invalid"`.

2. Full message of a base failure starts with a space symbol, e.g. `" the message"`.

**To Reproduce**

```ruby
require "i18n"
require "dry/validation"

I18n.backend.store_translations(
  :en,
  {
    dry_validation: {
      errors: {
        is_invalid: "is invalid"
      }
    }
  }
)

class MyContract < Dry::Validation::Contract
  config.messages.backend = :i18n

  params do
    required(:foo).value(:integer)
  end

  rule do
    base.failure("base failure here")
  end

  rule(:foo) do
    key.failure(:is_invalid)
  end
end

MyContract.new.call(foo: 42).errors(full: true).map(&:text)
#=> [" base failure here", " is invalid"]
# Expected:
#=> ["base failure here", "foo is invalid"]
```

**Expected behavior**

1. Full message of a key failure always starts with the key name, even if the key is not translated
2. Full message of a base failure includes just the message without the leading space symbol, e.g. `"the message"` instead of `" the message"`

**Environment**

ruby: 2.6.6
dry-validation: 1.5.2

stind:
Fixes #661

### DIFF
--- a/lib/dry/validation/messages/resolver.rb
+++ b/lib/dry/validation/messages/resolver.rb
@@ -1,10 +1,13 @@
 # frozen_string_literal: true
 
 require "dry/validation/message"
+require "dry/schema/message_compiler"
 
 module Dry
   module Validation
     module Messages
+      FULL_MESSAGE_WHITESPACE = Dry::Schema::MessageCompiler::FULL_MESSAGE_WHITESPACE
+
       # Resolve translated messages from failure arguments
       #
       # @api public
@@ -33,7 +36,7 @@ module Dry
           when Symbol
             Message[->(**opts) { message(message, path: path, tokens: tokens, **opts) }, path, meta]
           when String
-            Message[->(**opts) { [message_text(message, path, **opts), meta] }, path, meta]
+            Message[->(**opts) { [message_text(message, path: path, **opts), meta] }, path, meta]
           when Hash
             meta = message.dup
             text = meta.delete(:text) { |key|
@@ -50,18 +53,6 @@ module Dry
           end
         end
         alias_method :[], :call
-
-        # Resolve a message
-        #
-        # @return String
-        #
-        # @api public
-        def message_text(message, path, locale: nil, full: false, **opts)
-          keys = path.to_a.compact
-          msg_opts = EMPTY_HASH.merge(path: keys, locale: locale || messages.default_locale)
-
-          full ? "#{messages.rule(keys.last, msg_opts) || keys.last} #{message}" : message
-        end
 
         # Resolve a message
         #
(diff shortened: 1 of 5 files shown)


lib/dry/validation/messages/resolver.rb
```diff
@@ -90,12 +78,29 @@ def message(rule, tokens: EMPTY_HASH, locale: nil, full: false, path:)
...
         # rubocop:enable Metrics/AbcSize
 
         private
 
+        def message_text(text, path:, locale: nil, full: false)
+          return text unless full
+
+          key = key_text(path: path, locale: locale)
+
+          [key, text].compact.join(SPACE)
```
solnic, line 92: This should use a similar mechanism like in dry-schema, because since dry-schema 1.5.3, languages that don't use spaces between words are handled properly. See [this PR](https://github.com/dry-rb/dry-schema/pull/292/files) for more details. There's `Dry::Schema::MessageCompiler::FULL_MESSAGE_WHITESPACE` map that you could use here. In the future this should be encapsulated better but for now we can just re-use this constant.

lib/dry/validation/messages/resolver.rb
```diff
@@ -51,18 +51,6 @@ def call(message:, tokens:, path:, meta: EMPTY_HASH)
         end
         alias_method :[], :call
 
-        # Resolve a message
-        #
-        # @return String
-        #
-        # @api public
-        def message_text(message, path, locale: nil, full: false, **opts)
```
esparta, line 59: message_text was marked as public on the API documentation:

https://rubydoc.info/gems/dry-validation/1.5.3/Dry/Validation/Messages/Resolver#message_text-instance_method

For compatibility reason a changing the scope to private and their signature would potentially break systems using this internally.

lib/dry/validation/messages/resolver.rb
stind: Will do that.

lib/dry/validation/messages/resolver.rb
stind: That's a good point. I'll return the method back with the original signature, so the bug fix could be shipped in a patch version.

There is something to consider though:
- the interface is not consistent
`#message` and `#call` expect `path` as a keyword argument
`#message_text` expects it as a positional argument

- should `#message` and `#message_text` be public at all? Isn't it enough to have one public `#call` method?

lib/dry/validation/messages/resolver.rb
stind: Should it depend on dry-schema? I've added the same constant to the resolver to not depend on that dry-schema version, but maybe you had something else in mind?

lib/dry/validation/messages/resolver.rb
solnic: @stind these shouldn't be public, it's a mistake in the yard docs, so don't worry about it. I'm 100% sure nobody is using this.

lib/dry/validation/messages/resolver.rb
```diff
@@ -9,6 +9,15 @@ module Messages
...
       class Resolver
+        FULL_MESSAGE_WHITESPACE = Hash.new(" ").merge(
+          ja: "",
+          zh: "",
+          bn: "",
+          th: "",
+          lo: "",
+          my: "",
+        ).freeze
+
```
solnic, line 20: @stind please just use the constant from dry-schema

lib/dry/validation/messages/resolver.rb
```diff
@@ -9,6 +9,15 @@ module Messages
...
       # @api public
       class Resolver
+        FULL_MESSAGE_WHITESPACE = Hash.new(" ").merge(
+          ja: "",
+          zh: "",
+          bn: "",
+          th: "",
+          lo: "",
+          my: "",
+        ).freeze
```
solnic, line 19: @stind please just use the constant from dry-schema

PR #664, opened 2020-08-11
Evaluator#key? forwards key name argument

alassek:
When testing the presence of multiple, optional keys, it would be nice to be able to specify the key name to `key?` as you can with `values.key?`.

Since this merely delegates to that method, I can't see any reason why this omission was intentional. `#key_name` is preserved as the default argument, so this will not change any existing behavior.

[discussed in the forum here](https://discourse.dry-rb.org/t/better-consistency-for-key-method/1108/2)

### DIFF
--- a/docsite/source/rules.html.md
+++ b/docsite/source/rules.html.md
@@ -181,6 +181,23 @@ contract.call(email: 'jane@doe.org', login: 'jane', password: "").errors.to_h
 # => {:password=>["password is required"]}
 ```
 
+The `key?` method supports passing an explicit key name for rules that have multiple keys.
+
+```ruby
+class DistanceContract < Dry::Validation::Contract
+  schema do
+    optional(:kilometers).filled(:integer)
+    optional(:miles).filled(:integer)
+  end
+
+  rule(:kilometers, :miles) do
+    if key?(:kilometers) ^ key?(:miles)
+      base.failure("must only contain one of: kilometers, miles")
+    end
+  end
+end
+```
+
 ### Checking for previous errors
 
 Sometimes you may be interested in adding an error when some other error has happened.
--- a/lib/dry/validation/evaluator.rb
+++ b/lib/dry/validation/evaluator.rb
@@ -155,16 +155,23 @@ module Dry
       #
       # This is useful when dealing with rules for optional keys
       #
-      # @example
+      # @example use the default key name
       #   rule(:age) do
       #     key.failure(:invalid) if key? && value < 18
+      #   end
+      #
+      # @example specify the key name
+      #   rule(:start_date, :end_date) do
+      #     if key?(:start_date) && !key?(:end_date)
+      #       key(:end_date).failure("must provide an end_date with start_date")
+      #     end
       #   end
       #
       # @return [Boolean]
       #
       # @api public
-      def key?
-        values.key?(key_name)
+      def key?(name = key_name)
+        values.key?(name)
       end
 
       # Check if there are any errors on the schema under the provided path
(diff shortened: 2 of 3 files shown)


docsite/source/rules.html.md
```diff
@@ -181,6 +181,23 @@ contract.call(email: 'jane@doe.org', login: 'jane', password: "").errors.to_h
 # => {:password=>["password is required"]}
 ```
 
+The `key?` method supports passing an explicit key name for rules that have multiple keys.
+
+```ruby
+class DistanceContract < Dry::Validation::Contract
```
esparta, line 187: Can we have a spec at contract level (/spec/integration/contract/) mimicking this same scenario where we are asking of an specific keys?

lib/dry/validation/evaluator.rb
```diff
@@ -155,16 +155,23 @@ def value
...
+      #       key(:end_date).failure("must provide an end_date with start_date")
+      #     end
+      #   end
+      #
       # @return [Boolean]
       #
       # @api public
-      def key?
-        values.key?(key_name)
+      def key?(name = key_name)
```
solnic, line 173: I think we want to support the same type of args as `Values#key?`, WDYT? Values supports symbols, arrays, path objects etc. so this could be something like:

```ruby
def key?(*args)
  args.empty? values.key?(key_name) : values.key?(*args)
end
```

alassek: I was intending this to work with `Schema::Path`, yes.

The [`Values` implementation doensn't rely on splatting](https://github.com/dry-rb/dry-validation/blob/master/lib/dry/validation/values.rb#L64-L80) to work, is that necessary here?
```ruby
      def key?(key, hash = data)
        return hash.key?(key) if key.is_a?(Symbol)

        Schema::Path[key].reduce(hash) do |a, e|
          if e.is_a?(Array)
            result = e.all? { |k| key?(k, a) }
            return result
          elsif e.is_a?(Symbol) && a.is_a?(Array)
            return false
          else
            return false unless a.is_a?(Array) ? (e >= 0 && e < a.size) : a.key?(e)
          end
          a[e]
        end

        true
      end
```

solnic: @alassek ah right, doesn't have to be `*args`, good point!

docsite/source/rules.html.md
alassek: Sure, I think the test coverage is pretty comprehensive already but that would probably better communicate the intent of this.

docsite/source/rules.html.md
```diff
@@ -181,6 +181,23 @@ contract.call(email: 'jane@doe.org', login: 'jane', password: "").errors.to_h
...
+  schema do
+    optional(:kilometers).filled(:integer)
+    optional(:miles).filled(:integer)
```
solnic, line 190:
```suggestion
    optional(:miles).value(:integer)
```

docsite/source/rules.html.md
```diff
@@ -181,6 +181,23 @@ contract.call(email: 'jane@doe.org', login: 'jane', password: "").errors.to_h
 # => {:password=>["password is required"]}
 ```
 
+The `key?` method supports passing an explicit key name for rules that have multiple keys.
+
+```ruby
+class DistanceContract < Dry::Validation::Contract
+  schema do
+    optional(:kilometers).filled(:integer)
```
solnic, line 189:
```suggestion
    optional(:kilometers).value(:integer)
```

docsite/source/rules.html.md
solnic: I merged it as-is because having this example in the docs already covers the intended usage and the implementation is so simple that the unit test will suffice.

solnic:
@alassek no worries!

PR #665, opened 2020-08-19
Support per-key custom errors for array paths

### LINKED ISSUE #663: Support per-key custom errors for array paths
Given the following contract:

```ruby
config.messages.backend = :i18n
config.message.namespace = :my_contract

params do
  required(:people).filled(:array)
end

rule(:people).each do
   key.failure(:invalid) if person_invalid?
end
```

it should be possible to define the message for `:invalid` under `:people` as follows:

```yaml
en:
  dry_validation:
    errors:
      my_contract:
        rules:
          people:
            invalid: 'Something'
```

However, currently, this gives an error that the `:invalid` key can't be found under `people.0`. Instead, you need to define the translations as such:

```yaml
en:
  dry_validation:
    errors:
      my_contract:
        invalid: 'Something'
```

_Issue originally reported [on Discourse](https://discourse.dry-rb.org/t/i18n-paths-for-array-rules/1097)_

- Affects my production application: **NO**
- Ruby version: 2.6.5

tadeusz-niemiec:
Resolves #663

This allows using custom errors by a key without an index when used in .each.
I guess there should be a better way to check if this is related to .each than by checking the keys array size, let me know if you have any ideas.

### DIFF
--- a/lib/dry/validation/messages/resolver.rb
+++ b/lib/dry/validation/messages/resolver.rb
@@ -69,6 +69,11 @@ module Dry
             template, meta = messages[rule, msg_opts.merge(path: keys.last)] unless template
           end
 
+          if !template && keys.size > 1
+            non_index_keys = keys.reject { |k| k.is_a?(Integer) }
+            template, meta = messages[rule, msg_opts.merge(path: non_index_keys.join(DOT))]
+          end
+
           unless template
             raise MissingMessageError, <<~STR
               Message template for #{rule.inspect} under #{keys.join(DOT).inspect} was not found
--- a/spec/fixtures/messages/errors.en.yml
+++ b/spec/fixtures/messages/errors.en.yml
@@ -17,3 +17,5 @@ en:
           wrong: "not right"
         age:
           toddler: "should be included in %{list}"
+        small_integers:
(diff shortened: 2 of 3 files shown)


lib/dry/validation/messages/resolver.rb
```diff
@@ -69,6 +69,12 @@ def message(rule, tokens: EMPTY_HASH, locale: nil, full: false, path:)
             template, meta = messages[rule, msg_opts.merge(path: keys.last)] unless template
           end
 
+          if !template && keys.size > 1
+            non_index_keys = []
+            keys.each { |k| non_index_keys << k unless k.is_a?(Numeric) }
```
flash-gordon, line 74: these two lines are `non_index_keys = keys.reject { |k| k.is_a?(Integer) }`

tadeusz-niemiec: lol, I tried to use `filter` but it was unavailable in the 2.4 version and I totally forgot about `reject`, thanks :)

flash-gordon: ruby's version of `filter` is `select` which IMO makes more sense because it's not obvious if `filter` filters _in_ or _out_

PR #668, opened 2020-09-03
Update rules.html.md

fig:
I'm fairly sure the logic was wrong here, but please double check.

### DIFF
--- a/docsite/source/rules.html.md
+++ b/docsite/source/rules.html.md
@@ -191,7 +191,7 @@ class DistanceContract < Dry::Validation::Contract
   end
 
   rule(:kilometers, :miles) do
-    if key?(:kilometers) ^ key?(:miles)
+    if key?(:kilometers) && key?(:miles)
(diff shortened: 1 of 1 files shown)


docsite/source/rules.html.md
```diff
@@ -191,7 +191,7 @@ class DistanceContract < Dry::Validation::Contract
   end
 
   rule(:kilometers, :miles) do
-    if key?(:kilometers) ^ key?(:miles)
+    if key?(:kilometers) && key?(:miles)
```
flash-gordon, line 194: `unless key?(:kilometers) ^ key?(:miles)` would make more sense IMO :)

fig: I did consider whether that might be the intention, but then the message wouldn't make sense.
This class looks like it actually requires 2 rules -

- "must only contain one of: kilometers, miles"
- "must contain one of: kilometers, miles"

PR #671, opened 2020-09-16
correct the code in some of the examples

dreig:
Some of the examples used outdated code, I think from before dry-schema was separated from dry-validation (?)


solnic:
Thanks, and yes these examples were "ancient"

PR #672, opened 2020-09-17
Fix Undefined method `key?' for nil:NilClass error.

### LINKED ISSUE #670: Rule#each. Undefined method `key?' for nil:NilClass
**Describe the bug**

After calling contract with nested hash when the root key is nil and rule with `.each` applied the error occurs
```
NoMethodError:
   undefined method `key?' for nil:NilClass
```
**To Reproduce**
Giving a simple contract like this one:

```
class Index < Dry::Validation::Contract
  params do
    optional(:filter).hash do
      optional(:user_ids).array(:integer)
    end
   end

  rule(filter: :user_ids).each do
    key.failure('Not found') unless User.exists?(value)
  end
end

Index.new.call(filter: nil)
```
**Expected behavior**

**Expected behavior**
The contract does not raise any errors.

**Your environment**
- dry-validation 1.5.6
- Ruby 2.6.5

alexxty7:
Fixes #670

### DIFF
--- a/lib/dry/validation/values.rb
+++ b/lib/dry/validation/values.rb
@@ -70,6 +70,8 @@ module Dry
             return result
           elsif e.is_a?(Symbol) && a.is_a?(Array)
             return false
+          elsif a.is_a?(String) || a.nil?
+            return false
           else
             return false unless a.is_a?(Array) ? (e >= 0 && e < a.size) : a.key?(e)
           end
--- a/lib/dry/validation/values.rb
+++ b/lib/dry/validation/values.rb
@@ -70,7 +70,9 @@ module Dry
             return result
           elsif e.is_a?(Symbol) && a.is_a?(Array)
             return false
-          elsif a.is_a?(String) || a.nil?
+          elsif a.nil?
+            return false
+          elsif a.is_a?(String)
             return false
           else
             return false unless a.is_a?(Array) ? (e >= 0 && e < a.size) : a.key?(e)
--- a/spec/unit/values_spec.rb
+++ b/spec/unit/values_spec.rb
@@ -8,7 +8,13 @@ RSpec.describe Dry::Validation::Values do
   end
 
   let(:data) do
-    {name: "Jane", address: {city: "Paris", geo: {lat: 1, lon: 2}}, phones: [123, 431]}
+    {
+      name: "Jane",
+      address: {city: "Paris", geo: {lat: 1, lon: 2}},
+      phones: [123, 431],
+      billing_address: nil,
+      card: ""
+    }
   end
 
   describe "#[]" do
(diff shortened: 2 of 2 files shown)


lib/dry/validation/values.rb
```diff
@@ -70,6 +70,8 @@ def key?(key, hash = data)
             return result
           elsif e.is_a?(Symbol) && a.is_a?(Array)
             return false
+          elsif a.is_a?(String) || a.nil?
+            return false
```
solnic, line 74: Why `a.is_a?(String)` was needed? It shouldn't be, at this point all keys should be symbolized.

solnic (changes requested):
Thanks for addressing this! I need to clarify one thing before this can be merged, please see my question in the review.

lib/dry/validation/values.rb
alexxty7: It for the case when root value is a string. I have checked and it is not symbolized.
It covered by this test case
```
  let(:data) do
    {
      name: "Jane",
      address: {city: "Paris", geo: {lat: 1, lon: 2}},
      phones: [123, 431],
      billing_address: nil,
      card: ""
    }
  end

    it "returns false when a nested key is not present and root key is string" do
      expect(values.key?([:card, :not_here])).to be(false)
    end

```

alexxty7: Maybe I should check for valid types instead?

```
          elsif !(a.is_a?(Array) || a.is_a?(Hash))
            return false
```

solnic: @alexxty7 aahh that's right, this can happen! I think it should be handled separately though, so:

```ruby
# ...
elsif a.nil?
  return false
elsif a.is_a?(String)
  return false
# ...
end
```

alexxty7: @solnic updated

solnic (approved):
Thank you!

TimoMoss:
hello, 
I  have a similar problem when I use a rule for each element of an array. 
dry-validation  '~> 1.4'

```
ValidateParams < Dry::Validation::Contract
    schema do
        required(:data).hash do
          required(:attributes).hash do
            required(:items).array(:string)
         end
       end
    end

    rule(%i[data attributes items]).each do
       message = 'invalid numeric string'
       key(keys.flatten).failure(message) if value&.!~ /^(\d)+$/
    end
 end
```

```
RSpec.describe ValidateParams do
  let(:validation) { described_class.new }

  let(:params) do
    {
        data: {
            attributes: {
               items: ['123', '123', '123']
            }
        }
    }
  end

  it 'looks like failure' do
    expect(validation.call(params)).to be_failure
  end
end
```
result 
```
>>  Failure/Error: expect(validation.call(params)).to be_failure
     NoMethodError:
       undefined method `key?' for nil:NilClass
```

at the moment I have it solved by:

      ```
      rule(%i[data attributes items]) do
         value.each_with_index do |v, i|
           message = 'invalid numeric string'
           key(keys.flatten << i).failure(message) if v&.!~ /^(\d)+$/
        end
      end
```
But it's definitely a trick.
I'd highly appreciate it if anybody explains how can I make it work without such tricks.   
Many thanks in advance

PR #673, opened 2020-10-06
adds support of explict argument for #rule_error? method

moofkit:
Closing https://github.com/dry-rb/dry-validation/issues/658

Adds support for `#rule_error?` for checking that other rule has error
### Example
```ruby
class PersonContract < Dry::Validation::Contract
  schema do
    required(:email).filled(:string)
    required(:name).filled(:string)
  end

  rule(:name) do
    key.failure('name rule error')
  end

  rule(:email) do
    key.failure('email rule error') if rule_error?(:name)
  end
end

PersonContract.new.call(email: 'bar', name: 'foo').errors.to_h
# {name: ['name rule error'], email: ['email rule error']}
```

### DIFF
--- a/lib/dry/validation/evaluator.rb
+++ b/lib/dry/validation/evaluator.rb
@@ -187,11 +187,17 @@ module Dry
 
       # Check if there are any errors on the current rule
       #
+      # @param path [Symbol, String, Array] A Path-compatible spec
+      #
       # @return [Boolean]
       #
       # @api public
-      def rule_error?
-        !key(path).empty?
+      def rule_error?(path = nil)
+        if path.nil?
+          !key(self.path).empty?
+        else
+          result.error?(path)
+        end
       end
 
       # @api private
--- a/lib/dry/validation/evaluator.rb
+++ b/lib/dry/validation/evaluator.rb
@@ -196,7 +196,7 @@ module Dry
         if path.nil?
           !key(self.path).empty?
         else
-          result.error?(path)
+          result.rule_error?(path)
         end
       end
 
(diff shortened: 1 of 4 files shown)


lib/dry/validation/evaluator.rb
```diff
@@ -187,11 +187,17 @@ def schema_error?(path)
...
       # @return [Boolean]
       #
       # @api public
-      def rule_error?
-        !key(path).empty?
+      def rule_error?(path = nil)
+        if path.nil?
+          !key(self.path).empty?
+        else
+          result.error?(path)
```
solnic, line 199: This checks all errors, not just rule errors, so I'm afraid this doesn't do what we want. One option to make this work would be to add `Result#rule_error?` that would filter out schema errors and look only at rule errors.

solnic (changes requested):

lib/dry/validation/evaluator.rb
moofkit: The problem here that schema and rule errors are mixed in `Result#errors` set.
So I see two possible solutions:
1. Filter schema errors keys from errors set and searching in that subset.
2. Check that there are no schema error with provided key and then look up `Result#errors`.

The first approach seems to me is a little more straightforward.
Because in second one you need to know that rule dependent on schema and does not evaluate if there are some error.
But the second one was easier to implement 😄 

If second approach is not ok for you I will rewrite it.

moofkit:
@solnic thanks for your review!

I have fixed issue. Can you please have a look one more time? Or I need to rewrite it with another approach?

solnic (approved):

solnic:
@moofkit hey sorry for the delay, I was on a sick leave. Thanks for working on this PR. As you can see I just merged it and it'll be included in the 1.6.0 release.

moofkit:
@solnic thanks! Hope your are well for now!

Issue #356, opened 2017-08-11
TypeError when trying to access messages on result


jmrepetti:
Hello, I'm trying to access the messages on the result and it fail when I define the schema using the block syntax for a required filled attribute. 

I'm using **ruby 2.4.1p111** On **OSX Darwin 16.7.0 Darwin Kernel Version 16.7.0**

I didn't have much time to debug it. Below there's and example of the problem, the schemas are a bit different. On the failing one, I use a logic predicate to allow int or string. 

It's me the problem?

Working schema:

```ruby
work = Dry::Validation.Schema do
  required(:nested_schema).schema do
    required(:code).filled(:str?)
  end
end

work.({}).messages
=> {:nested_schema=>["is missing"]}
```

Failing schema:

```ruby
doesnt_work = Dry::Validation.Schema do
  required(:nested_schema).schema do
    required(:code).filled { str? | int? }
  end
end      

doesnt_work.({}).messages
TypeError: no implicit conversion of Symbol into Integer
	from ($HOME)/Development/dry-validation/lib/dry/validation/message_set.rb:113:in `[]'
	from ($HOME)/Development/dry-validation/lib/dry/validation/message_set.rb:113:in `block in initialize_placeholders!'
	from ($HOME)/Development/dry-validation/lib/dry/validation/message_set.rb:106:in `each'
	from ($HOME)/Development/dry-validation/lib/dry/validation/message_set.rb:106:in `reduce'
	from ($HOME)/Development/dry-validation/lib/dry/validation/message_set.rb:106:in `initialize_placeholders!'
	from ($HOME)/Development/dry-validation/lib/dry/validation/message_set.rb:27:in `initialize'
	from ($HOME)/Development/dry-validation/lib/dry/validation/message_set.rb:16:in `new'
	from ($HOME)/Development/dry-validation/lib/dry/validation/message_set.rb:16:in `[]'
	from ($HOME)/Development/dry-validation/lib/dry/validation/message_compiler.rb:36:in `call'
	from ($HOME)/Development/dry-validation/lib/dry/validation/result.rb:53:in `message_set'
	from ($HOME)/Development/dry-validation/lib/dry/validation/result.rb:41:in `messages'
	from (irb):6
```

**node** variable inside the while block result in empty array and it fail when trying to do **node[key]**

```ruby
#file dry-validation/lib/dry/validation/message_set.r:105
      def initialize_placeholders!
        @placeholders = paths.reduce({}) do |hash, path|
[...]

PR #686, opened 2021-05-24
Update for latest dry-configurable setting API

timriley:


solnic (commented):
I think I want to make it work with both versions until 2.0.0. Thoughts?

timriley:
@solnic good thought, and I think you're right, particularly for the post-1.0 gems of ours. Do you have any thoughts about how to go about it?

timriley:
@solnic I like what you did over in [`0471911` (dry-schema #356)](https://github.com/dry-rb/dry-schema/pull/356/commits/0471911aa95a184f594cfded07f762becad09489) — would you be happy with that approach here too?

solnic:
@timriley yes sorry that's what I meant 🙂 the only "issue" is that we would have to ensure the new dry-validation depends on the new dry-schema too...unless you want to override `setting` here as well?

timriley:
@solnic Yeah, I think we'd want to put this shim directly in every gem that uses such an approach

solnic:
@timriley done here c8ad125

timriley:
Thanks to the improved backwards compatibility in the upcoming dry-configurable deprecations, this one no longer needed any kind of additional shim.

PR #690, opened 2021-09-09
Add new method to check whether base_errors have already occured

wuarmin:
Hey,

I posted a question at [discourse](https://discourse.dry-rb.org/t/is-there-a-api-to-check-if-base-error-happened/1332). I need an api to check if a base_error has already occured. This PR would add the method `base_rule_error?`.

WDYT?
best regards


solnic:
Thanks and I'm sorry it took me so long to get to it. I'll push 1.8.0 with this feature included.

PR #695, opened 2021-12-04
improve: order of nested test

JuanVqz:


solnic:
Thanks but what is exactly improved here?

JuanVqz:
I was extracting the following block code in a `before each` I thought it may be an improvement.

```ruby
      contract_class.rule(:tags) do
        key.failure("must have at least 1 element") unless value.size > 0
      end
```

spec/unit/rule_spec.rb
```diff
@@ -14,21 +14,19 @@
       end
     end
 
-    it "allows specifying a rule for array elements" do
+    before :each do
```
solnic, line 17:
```suggestion
    before do
```

solnic (changes requested):

PR #696, opened 2021-12-04
improve: values subject spec

JuanVqz:


solnic:
Thanks but in this case it's not really that helpful. In general, described_class is actually a private RSpec API. Personally I'm not a fan of using it.

PR #697, opened 2021-12-04
improve: Dry::Validation::Result test

JuanVqz:
A little bit of DRY on test


solnic:
Thanks but this type of indirection isn't really helpful in case of a spec

PR #698, opened 2021-12-04
improve: pattern matching test

JuanVqz:
replace single quotes to double quotes.


solnic (changes requested):
Please remove the reverted commit and force push here so that I can merge it in without the extra commit

JuanVqz:
> Please remove the reverted commit and force push here so that I can merge it in without the extra commit

@solnic done!

PR #706, opened 2022-03-03
Raise an InvalidKeyErrors on substring of valid keys

### LINKED ISSUE #705: Incorrect validation of rule keys substring of schema keys
## Describe the bug

Contract#rule does not reject invalid keys if they are substring of valid keys

## To Reproduce

<details>
<summary>
reproduction script
</summary>

```ruby
require 'bundler/inline'
gemfile(true) do
  source 'https://rubygems.org'

  gem 'dry-validation'
end

class Test < Dry::Validation::Contract
  schema do
    required(:tags).array(:hash) do
      required(:name).filled(:string)
    end
  end
  
  rule(:tag) do # This should raise an InvalidKeysError
    key.failure('must have at least 1 element') if value.empty?
  end
end
```

</details>

## Expected behavior

Contract#rule should raise an InvalidKeysError

## My environment

- Affects my production application: NO
- Ruby version: 2.7
- OS: macOS

MatElGran:
Fixes #705

I'm not completely sure of the fix, but it's probably good enough
as a starting point

### DIFF
--- a/lib/dry/validation/contract/class_interface.rb
+++ b/lib/dry/validation/contract/class_interface.rb
@@ -158,13 +158,13 @@ module Dry
           valid_paths = key_map.to_dot_notation
           key_paths = key_paths(keys)
 
-          invalid_keys = key_paths.map { |(key, path)|
+          invalid_keys = key_paths.filter_map { |(key, path)|
             unless valid_paths.any? { |vp|
-                     vp == path || vp.include?("#{path}.") || vp.include?("#{path}[]")
+                     vp == path || vp.start_with?("#{path}.", "#{path}[]")
                    }
               key
             end
-          }.compact.uniq
+          }.uniq
 
           return if invalid_keys.empty?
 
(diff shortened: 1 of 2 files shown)


lib/dry/validation/contract/class_interface.rb
```diff
@@ -159,7 +159,9 @@ def ensure_valid_keys(*keys)
           key_paths = key_paths(keys)
 
           invalid_keys = key_paths.map { |(key, path)|
-            unless valid_paths.any? { |vp| vp.include?(path) || vp.include?("#{path}[]") }
+            unless valid_paths.any? { |vp|
+                     vp == path || vp.include?("#{path}.") || vp.include?("#{path}[]")
+                   }
```
solnic, line 164: I reckon we need to use a regexp here, otherwise `"footag.bar"` would have the same issue, right?

solnic (changes requested):

lib/dry/validation/contract/class_interface.rb
MatElGran: Oh yes, you're right, I missed that case. Wouldn't `String#start_with?` also fix it while being more legible though?

MatElGran: Well, it does indeed fix it, as for the more legible part, I'll let you decide ;)

solnic (approved):

solnic:
Thanks man ❤️ I'm gonna release it in 1.8.1

PR #714, opened 2022-09-22
Update config in configure blocks only

timriley:
Update all usages of dry-configurable-provided `config` so that any changes are made inside `configure` blocks.

This prepares for the changes in https://github.com/dry-rb/dry-configurable/pull/140.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gemspec
 if ENV["DRY_CONFIGURABLE_FROM_MAIN"].eql?("true")
   gem "dry-configurable", github: "dry-rb/dry-configurable", branch: "main"
 end
+gem "dry-core", github: "dry-rb/dry-core"
 
 if ENV["DRY_SCHEMA_FROM_MAIN"].eql?("true")
   gem "dry-schema", github: "dry-rb/dry-schema", branch: "main"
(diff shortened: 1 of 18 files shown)


solnic:
I'm sorry but can this only include configure change? I don't want zeitwerk stuff to be in the main branch just yet. It's not ready (dry-core is not released) and it already caused two invalid releases of dry-schema (which is why I reverted zeitwerk changes in its main branch).

timriley:
@solnic Yep, I can do this. Weirdly, dry-configurable in the `main` branch has already been updated to expect the Zeitwerk-enabled dry-core, which is why all these gems I'm testing with dry-configurable also need dry-core updated.

Do we want to keep zeitwerk-enabled dry-core? What is our plan for Zeitwerk in our other gems pre-1.0? Because I want to bring everything to 1.0 (at least as release candidates) basically in the next month.

timriley:
What I'll do for starters is just revert the commit in dry-configurable's main branch that expects the zeitwerk-enabled dry-core.

timriley:
@solnic FYI, with https://github.com/dry-rb/dry-configurable/pull/142 now merged, I've updated this branch so it is no longer changing anything related to requiring code.

PR #737, opened 2024-11-21
Add funding_uri to gemspec

andrew:
I noticed that the project already has a FUNDING.yml, this PR adds `funding_uri` to the gemspec to help increase visibility using the `bundle fund` command in Bundler.

Happy to open similar PRs for all the dry-rb gems if this is accepted.


timriley:
Thanks for looking out for this, @Andrew! It was really nice to see this PR come my way :)

So here's a twist: all the gemspecs across the dry-rb projects are auto-generated from this file: https://github.com/dry-rb/template-gem/blob/main/gemspec.erb

So if you put the funding_uri there, it should flow on to all dry-rb gems at once. Would you be up for taking care of that? 😄 Thanks!

